PR #4, opened 2024-08-19
Initial stab at GraphQL client

stefan-mysten:
Added an initial implementation of the GraphQL client with some basic API.

Would love to hear feedback on 
* handling errors
* how to test this - snapshots? having like a little local network test cluster and then making TXs and using the API?


crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,568 @@
...
+    }
+
+    fn new(page_info: PageInfo, nodes: Vec<T>) -> Self {
+        Self { page_info, nodes }
+    }
+}
+
+/// The SuiClient is a GraphQL client for interacting with the Sui blockchain.
+/// By default, it uses Reqwest as the HTTP client.
+pub struct SuiClient {
```
bmwill, line 89: Lets rename this to just `Client` and when we use the reqwest client we can just name it via its fully qualified type `reqwest::Client`. The additional "Sui" part of the name doesn't convey any additional information so it could be dropped. That and given namespacing we don't need to worry about this being named "Client" and something else being named "Client".

We could prepend it with GraphQlClient or some such, but even that isn't strictly necessary since the package itself is named sui-graphql-client.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,568 @@
...
+}
+
+impl<T> Page<T> {
+    pub fn page_info(&self) -> &PageInfo {
+        &self.page_info
+    }
+
+    pub fn nodes(&self) -> &Vec<T> {
+        &self.nodes
+    }
```
bmwill, line 80:
```suggestion
    pub fn nodes(&self) -> &[T] {
        &self.nodes
    }
```

crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,568 @@
...
+impl Default for SuiClient {
+    /// Create a new [`SuiClient`] with testnet as the default GraphQL server.
+    fn default() -> Self {
+        Self {
+            rpc: TESTNET_HOST.to_string(),
+            rpc_version: None,
+            inner: ReqwestHttpClient::new(),
+        }
+    }
+}
```
bmwill, line 107: I think we should avoid having a Default impl for this Client. Its not clear what the default is without reading the docs (which is fine) but its probably clearer to have specific constructors for each of the common networks + `new` that just takes the url (which the other ones will use internally for construction)

crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,568 @@
...
+    pub fn set_testnet(&mut self) -> &Self {
+        self.rpc = TESTNET_HOST.to_string();
+        self
+    }
+
+    /// Set GraphQL client to mainnet.
+    pub fn set_mainnet(&mut self) -> &Self {
+        self.rpc = MAINNET_HOST.to_string();
+        self
+    }
```
bmwill, line 162: Generally methods that are of the form `set_*` won't return anything. It sort of seems like you're half-way between wanting a Builder pattern, in which case you'd want to return either `Self` or `&mut Self` so that further changes could be made via chaining.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,568 @@
...
+    /// intended to be used with custom queries.
+    pub async fn run_query<
+        T: serde::de::DeserializeOwned + Send,
+        V: serde::Serialize + Send + std::marker::Sync,
+    >(
+        &self,
+        operation: &Operation<T, V>,
+    ) -> Result<GraphQlResponse<T>> {
+        self.inner.post(&self.url(), operation).await
+    }
```
bmwill, line 175:
```suggestion
    pub async fn run_query<T, V>(
        &self,
        operation: &Operation<T, V>,
    ) -> Result<GraphQlResponse<T>>
    where
        T: serde::de::DeserializeOwned,
        V: serde::Serialize,
    {
        self.inner.post(&self.url(), operation).await
    }
```

lets use a where clause, its a bit easier to read. And are the send/Sync reqs coming from Operation? I don't think they're strictly necessary.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,568 @@
...
+    ChainIdentifierQuery, CheckpointArgs, CheckpointId, CheckpointQuery, CoinMetadata,
+    CoinMetadataArgs, EpochSummaryArgs, EpochSummaryQuery, EventFilter, ObjectFilter, ObjectQuery,
+    ObjectQueryArgs, PageInfo, ProtocolConfigQuery, ProtocolConfigs, ProtocolVersionArgs,
+    ServiceConfig, ServiceConfigQuery, TransactionBlockArgs, TransactionBlockQuery, Uint53,
+};
+use sui_types::types::{
+    Address, CheckpointSequenceNumber, CheckpointSummary, Event, Object, SignedTransaction,
+};
+
+use anyhow::{ensure, Error, Result};
```
bmwill, line 19: As a first pass at implementing this library using anyhow is a great choice! Once we've more fleshed things out we should look at using a custom error type to allow for a bit more flexibility/context. But thats something that we should punt till later

crates/sui-graphql-client/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+reqwest = { version = "0.12", features = ["json", "blocking"] }
+sui-types = { package= "sui-sdk", path = "../sui-sdk", features = ["serde"] }
+tokio = { version = "1.39.2", features = ["full"] }
+
+[dev-dependencies]
+anyhow = "1.0.86"
+cynic = { version = "3.7.3" }
+serde = { version = "1.0.208", features = ["serde_derive"] }
+serde_json = "1.0.70"
+tokio = { version = "1.39.2", features = ["full"] }
```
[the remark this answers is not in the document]
stefan-mysten: Did not know that, thanks. This is not needed after removing the `surf` example.

crates/sui-graphql-client/src/lib.rs
stefan-mysten: I do not know to be honest, compiler is complaining so I added them as suggested by the compiler.

crates/sui-graphql-client/src/lib.rs
stefan-mysten: Sounds good!

crates/sui-graphql-client/src/lib.rs
stefan-mysten: That sounds better! I was undecided on how to go about it.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,584 @@
...
+    /// Create a new GraphQL client connected to the `devnet` GraphQL server: {DEVNET_HOST}.
+    pub fn new_devnet() -> Result<Self, Error> {
+        Self::new(DEVNET_HOST)
+    }
+
+    /// Create a new GraphQL client connected to the `localhost` GraphQL server:
+    /// {DEFAULT_LOCAL_HOST}.
+    pub fn new_localhost(&mut self) -> Result<Self, Error> {
+        Self::new(DEFAULT_LOCAL_HOST)
+    }
```
bmwill, line 99: All of these should not fail, and if they do then its a library error. Lets unwrap and avoid the unnecessary error.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,584 @@
...
+
+    /// Return the URL for the GraphQL GraphQL client after checking if the version is set.
+    fn url(&self) -> String {
+        if let Some(version) = &self.rpc_version {
+            format!("{}/{}", self.rpc, version)
+        } else {
+            self.rpc.to_string()
+        }
+    }
```
bmwill, line 125: I would return a borrowed reference to the url vs allocating a string each time here. if the version is a required part of the URL, then it should always be a part of the url that is being used

crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,584 @@
...
+}
+
+impl Client {
+    // ===========================================================================
+    // Client Misc API
+    // ===========================================================================
+
+    /// Create a new GraphQL client with the provided server address.
+    pub fn new(server: &str) -> Result<Self, Error> {
+        reqwest::Url::parse(server).map_err(|_| anyhow!("Invalid URL: {}", server))?;
```
bmwill, line 70: if we're parsing the URL here, lets store the URL instead of a string.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,584 @@
...
+
+        let client = Client {
+            rpc: server.to_string(),
+            rpc_version: None,
+            inner: reqwest::Client::new(),
```
bmwill, line 75: could we add a default UserAgent header to the client? will make it easier to track in the future.

crates/sui-graphql-client/src/query_types/object.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+#[derive(cynic::QueryVariables, Debug)]
+pub struct ObjectsQueryArgs<'a> {
+    pub after: Option<&'a str>,
+    pub before: Option<&'a str>,
+    pub filter: Option<ObjectFilter<'a>>,
+    pub first: Option<i32>,
+    pub last: Option<i32>,
+}
```
bmwill, line 41: Are these types not already defined by the generated schema?

crates/sui-graphql-client/src/lib.rs
stefan-mysten: Added.

crates/sui-graphql-client/src/query_types/object.rs
stefan-mysten: I might misunderstand things, but I thought they are generated and only used when `build` is called on a query to verify this query against the generated schema. Not sure though, I might be totally off here but had the same question as you, and that's how I understood it needs to be done from the documentation.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+    }
+
+    /// Create a new GraphQL client connected to the `devnet` GraphQL server: {DEVNET_HOST}.
+    pub fn new_devnet() -> Self {
+        Self::new(DEVNET_HOST).expect("Invalid devnet URL")
+    }
+
+    /// Create a new GraphQL client connected to the `localhost` GraphQL server:
+    /// {DEFAULT_LOCAL_HOST}.
+    pub fn new_localhost(&mut self) -> Self {
```
tzakian, line 99: Don't need the `&mut self` here.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+        Self::new(DEFAULT_LOCAL_HOST).expect("Invalid localhost URL")
+    }
+
+    /// Set the version for the GraphQL GraphQL client. The default set version is stable.
```
tzakian, line 103: nit
```suggestion
    /// Set the version for the GraphQL client. The default set version is stable.
```

crates/sui-graphql-client/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+    ///
+    /// Use the `service_config` API to get the `available_versions` field to know which versions
+    /// are supported by this service.
+    pub fn set_version(&mut self, version: Option<&str>) -> Result<&mut Self, Error> {
```
tzakian, line 107: I'm a bit confused here and in `set_rpc_server` why this takes and returns `&mut self`/`&mut Self`. I'd expect either for this to be `&mut self`/`()` or `self`/`Self`. But maybe I'm missing something?

bmwill: Yeah the function signature is a little odd, but given the graphql version changes we can actually remove all of this!

stefan-mysten:
@bmwill did another pass to solve some issues. I think this is good to land before it gets too big, and then I will work on testing and more GQL support in smaller PRs to move faster.

crates/sui-graphql-client/src/query_types/checkpoint.rs
```diff
@@ -0,0 +1,132 @@
...
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(schema = "rpc", graphql_type = "GasCostSummary")]
+pub struct GasCostSummary {
+    pub computation_cost: Option<BigInt>,
+    pub non_refundable_storage_fee: Option<BigInt>,
+    pub storage_cost: Option<BigInt>,
+    pub storage_rebate: Option<BigInt>,
+}
+
+impl TryInto<CheckpointSummary> for Checkpoint {
```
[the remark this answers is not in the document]
stefan-mysten: Added - also have an issue for this in the project's tracker.

PR #8, opened 2024-09-19
[GraphQL Client] Add tests

stefan-mysten:
This simply tests that it can connect to a network and run and get back a request successfully, without any failures. This can catch if something changes in the GraphQL url, schema, BCS issues, etc.

### DIFF
--- a/crates/sui-graphql-client/src/lib.rs
+++ b/crates/sui-graphql-client/src/lib.rs
@@ -25,7 +25,7 @@ use cynic::{serde, GraphQlResponse, Operation, QueryBuilder};
 const MAINNET_HOST: &str = "https://sui-mainnet.mystenlabs.com/graphql";
 const TESTNET_HOST: &str = "https://sui-testnet.mystenlabs.com/graphql";
 const DEVNET_HOST: &str = "https://sui-devnet.mystenlabs.com/graphql";
-const DEFAULT_LOCAL_HOST: &str = "http://localhost:9125/graphql";
+const DEFAULT_LOCAL_HOST: &str = "http://127.0.0.1:9125/graphql";
 static USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"),);
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


crates/sui-graphql-client/src/lib.rs
```diff
@@ -221,7 +221,7 @@ impl Client {
         seq_num: Option<u64>,
     ) -> Result<Option<CheckpointSummary>, Error> {
         ensure!(
-            digest.is_some() != seq_num.is_some(),
+            !(digest.is_some() && seq_num.is_some()),
```
bmwill, line 224: ||?

crates/sui-graphql-client/src/lib.rs
```diff
@@ -566,4 +575,171 @@ mod tests {
...
+    async fn test_chain_id() {
+        for n in NETWORKS.iter() {
+            let client = Client::new(n).unwrap();
+            let chain_id = client.chain_id().await;
+            assert!(chain_id.is_ok());
+            if *n == "http://sui-mainnet.mystenlabs.com/graphql" {
+                assert!(chain_id.unwrap() == "35834a8a");
+            }
+        }
+    }
```
bmwill, line 589: Have an array of `[network, Option<Id>]` to check against

crates/sui-graphql-client/src/lib.rs
```diff
@@ -551,6 +559,7 @@ impl Client {
 #[cfg(test)]
 mod tests {
     use crate::{Client, DEFAULT_LOCAL_HOST, DEVNET_HOST, MAINNET_HOST, TESTNET_HOST};
+    const NETWORKS: [&str; 3] = [MAINNET_HOST, TESTNET_HOST, DEVNET_HOST];
```
bmwill, line 562: I'd probably say we should only test against Mainnet and testnet. devnet isn't intended to be reliable enough to warrant blocking ci on

crates/sui-graphql-client/src/lib.rs
```diff
@@ -25,7 +25,7 @@ use cynic::{serde, GraphQlResponse, Operation, QueryBuilder};
 const MAINNET_HOST: &str = "https://sui-mainnet.mystenlabs.com/graphql";
 const TESTNET_HOST: &str = "https://sui-testnet.mystenlabs.com/graphql";
 const DEVNET_HOST: &str = "https://sui-devnet.mystenlabs.com/graphql";
-const DEFAULT_LOCAL_HOST: &str = "http://localhost:9125/graphql";
+const DEFAULT_LOCAL_HOST: &str = "http://127.0.0.1:9125/graphql";
```
bmwill, line 28: why did this need to be changed?

stefan-mysten: Oops, no need.

crates/sui-graphql-client/src/lib.rs
stefan-mysten: Both `digest` and `seq_num` can be `None`.

bmwill: i guess the negative logic is tripping me up.

stefan-mysten: I cannot ever figure out how to work with `ensure`. I tried `||`, !(.. || ..)`, and the previous one, they all fail in one case or another.

PR #10, opened 2024-09-20
sui-sdk-crypto: introduce crate to handle signing and verifying

bmwill:


crates/sui-sdk-crypto/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+#[allow(unused)]
+mod secp256r1;
+
+pub trait SuiSigner {
+    fn sign_transaction(&self, transaction: &Transaction) -> Result<UserSignature, SignatureError>;
+    fn sign_personal_message(
+        &self,
+        message: &PersonalMessage<'_>,
+    ) -> Result<UserSignature, SignatureError>;
+}
```
jordanjennings-mysten, line 28: Do public keys normalize somehow? or would it be good to leave each pubkey() call out of the trait.

jordanjennings-mysten: can each impl of SuiSigner return the same type for `public_key()` or should they just be kept separate. On chain each address resolves to some `0x123...` which might be useful when using a signer as a common function.

crates/sui-sdk-crypto/Cargo.toml
```diff
@@ -0,0 +1,62 @@
...
+
+[dependencies]
+signature = "2.2"
+sui-sdk = { version = "0.0.0", path = "../sui-sdk", default-features = false }
+
+# RNG support
+rand_core = { version = "0.6.4", optional = true }
+
+# ed25519 support
+ed25519-dalek = { version = "2.1.1", optional = true }
```
jnaulty, line 36: noting that this is a different lib than fastcrypto uses:

- https://github.com/MystenLabs/fastcrypto/blob/3366c26a746b72707572c4f3f05915e20d5c16c2/fastcrypto/src/ed25519.rs#L79
- https://github.com/penumbra-zone/ed25519-consensus 

not quite sure what the implementation differences are--but worth investigating if using for production workloads.

crates/sui-sdk-crypto/Cargo.toml
```diff
@@ -0,0 +1,62 @@
...
+
+# secp256r1 support
+p256 = { version = "0.13.2", default-features = false, features = ["ecdsa", "std"], optional = true }
+
+# secp256k1 support
+k256 = { version = "0.13.4", default-features = false, features = ["ecdsa"], optional = true }
```
jnaulty, line 42: i believe similar story here too (fastcrypto uses a different lib)

- https://github.com/MystenLabs/fastcrypto/blob/main/fastcrypto/src/secp256k1/mod.rs#L36 
- https://github.com/rust-bitcoin/rust-secp256k1/

jnaulty (commented):
Looking at the crypto core dependencies--

I think it's worth mentioning somewhere "Why not use fastcrypto".

Recording the out-of-band answer I recieved was:
- size / feature restraints (keeping things minimal)
- WASM compatible (some fastcrypto libs don't work easily in wasm)


Using alternative libraries increases the nakamoto coefficient for signing/verifying utilities, which is a boost for resiliency in case of supply chain issues.

crates/sui-sdk-crypto/Cargo.toml
jnaulty: this library does have an audit though: https://www.nccgroup.com/us/research-blog/public-report-entropyrust-cryptography-review/

crates/sui-sdk-crypto/Cargo.toml
jnaulty: given this is a "A WIP from-scratch rust sdk for the sui blockchain", I think a note to audit/verify the integrity of this dependency should suffice

crates/sui-sdk-crypto/src/lib.rs
bmwill: So a signing API doesn't necessarily need to provide its public key info (Every signature in Sui will curry this information anyway). So are you more interested in ensuring that you can take a signer and easily ask/get the sui address that it corresponds to or do you want the pubkey info as it would possibly appear in a multisig address?

jordanjennings-mysten: Either way can be fine, it just might be nice to identify which signer you currently have. It's a common function in other signer interfaces.

bmwill: yeah of course, we can definitely expose that info

bmwill:
Merging this (since its large) and will add an updated readme as a follow up

PR #11, opened 2024-09-24
Add faucet API

stefan-mysten:


crates/sui-graphql-client/src/faucet.rs
```diff
@@ -0,0 +1,235 @@
...
+pub struct CoinInfo {
+    pub amount: u64,
+    pub id: ObjectId,
+    pub transfer_tx_digest: TransactionDigest,
+}
+
+impl FaucetClient {
+    /// Construct a new `FaucetClient` with the given `reqwest::Client`. Defaults to Testnet
+    /// faucet.
+    pub(crate) fn new(inner: reqwest::Client, rpc: &Url) -> Self {
```
bmwill, line 82: Lets change this to only take a url, the client can be created inside.

crates/sui-graphql-client/src/faucet.rs
```diff
@@ -0,0 +1,235 @@
...
+        let faucet_url = if rpc.as_str() == LOCAL_HOST {
+            Url::parse(FAUCET_LOCAL_HOST).expect("Invalid faucet URL")
+        } else if rpc.as_str() == DEVNET_HOST {
+            Url::parse(FAUCET_DEVNET_HOST).expect("Invalid faucet URL")
+        } else {
+            Url::parse(FAUCET_TESTNET_HOST).expect("Invalid faucet URL")
+        };
```
bmwill, line 89: Why are you trying to match the url against a network?

crates/sui-graphql-client/src/faucet.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub fn devnet(mut self) -> Self {
+        self.faucet_url = Url::parse(FAUCET_DEVNET_HOST).expect("Invalid faucet URL");
+        self
+    }
+
+    /// Set to testnet faucet.
+    pub fn testnet(mut self) -> Self {
+        self.faucet_url = Url::parse(FAUCET_TESTNET_HOST).expect("Invalid faucet URL");
+        self
+    }
```
bmwill, line 109: Lets remove these methods that mutate the client and instead have them be constructors for network specific faucets

crates/sui-graphql-client/src/faucet.rs
```diff
@@ -0,0 +1,235 @@
...
+                    .replace("gas", "status")
+                    .as_str(),
+            )
+            .expect("Invalid faucet status URL");
+
+            let poll_response = tokio::time::timeout(FAUCET_REQUEST_TIMEOUT, async {
+                let mut interval = tokio::time::interval(FAUCET_POLL_INTERVAL);
+                loop {
+                    interval.tick().await;
+                    eprint!("Polling faucet status...");
```
bmwill, line 179: lets remove the eprint lines

crates/sui-graphql-client/src/faucet.rs
```diff
@@ -0,0 +1,235 @@
...
+                    let req = self
+                        .request_status(request_id.clone(), status_url.clone())
+                        .await;
+
+                    if let Ok(Some(poll_response)) = req {
+                        eprintln!("response: {:?}", poll_response);
```
bmwill, line 185: here too

bmwill: or you can change them to tracing logs

crates/sui-graphql-client/src/faucet.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    /// Request gas from a custom faucet service.
+    pub async fn request_url(
+        &self,
+        address: Address,
+        url: Url,
+    ) -> Result<Option<String>, anyhow::Error> {
+        self.request_impl(address, &url).await
+    }
```
bmwill, line 220: We probably don't need this if people can construct a client that is bespoke to the url they care about

crates/sui-graphql-client/src/faucet.rs
```diff
@@ -0,0 +1,235 @@
...
+        &self,
+        id: String,
+        url: Url,
+    ) -> Result<Option<BatchSendStatus>, anyhow::Error> {
+        let url = format!("{}/{}", url, id);
+        let response = self.inner.get(&url).send().await?;
+        let json = response.json::<BatchStatusFaucetResponse>().await?;
+        Ok(json.status)
+    }
+}
```
bmwill, line 235: same here, use the internal stored url vs passing one in and take id by &str

crates/sui-graphql-client/src/faucet.rs
stefan-mysten: so not reuse the reqwest::Client we already have? What's the reasoning for having multiple Clients?

crates/sui-graphql-client/src/faucet.rs
stefan-mysten: So that you do not need to call `.devnet()` or set up specifically a faucet URL if the public nodes are used.

crates/sui-graphql-client/src/faucet.rs
stefan-mysten: this one is trickier depending on the faucet service (if they're not using our own service). Because you could have different paths for status and requesting a URL, so we'd either need to store both, or allow for the status url to be passed.

crates/sui-graphql-client/src/faucet.rs
stefan-mysten: basically you can do 
```rust
  let client = Client::new_devnet();
  let coin = client.faucet().request_and_wait("address").await?;
  ```

File: crates/sui-graphql-client/src/faucet.rs (hunk 7 above)
Comment (bmwill) [reply crates/sui-graphql-client/src/faucet.rs:235]: we need to standardize on *something* so our service, and how the routes are defined, is the purpose of this client is it not?

Comment (stefan-mysten) [reply crates/sui-graphql-client/src/faucet.rs:235]: Not sure I follow, you mean we should set the standard (e.g., now we have `v1/gas` and `v1/status`) and custom faucet services should follow it?

Comment (bmwill) [reply crates/sui-graphql-client/src/faucet.rs:235]: yes exactly. if it doesn't follow that standard then this faucet implementation won't work either way

File: crates/sui-graphql-client/src/faucet.rs (hunk 2 above)
Comment (bmwill) [reply crates/sui-graphql-client/src/faucet.rs:89]: Then this should be done from the GraphQL client not the faucet client. its very confusing to pass something in and have it reject what you gave it and instead do something completely different.

Comment (bmwill) [reply crates/sui-graphql-client/src/faucet.rs:89]: I'd also argue that you shouldn't be able to go from graphql client -> faucet client. It conflates the two together and they could be very confusing if you're not using a whitelisted network address. It would be much clearer to require someone build a graphql client and a faucet client separately

File: crates/sui-graphql-client/src/faucet.rs (hunk 1 above)
Comment (stefan-mysten) [reply crates/sui-graphql-client/src/faucet.rs:82]: Maybe in this case it's not that performance critical, but I thought reusing the Client is ok as it is already behind an Arc.

File: crates/sui-graphql-client/src/faucet.rs (hunk 7 above)
Comment (stefan-mysten) [reply crates/sui-graphql-client/src/faucet.rs]: Fixed

File: crates/sui-graphql-client/src/faucet.rs (hunk 4 above)
Comment (stefan-mysten) [reply crates/sui-graphql-client/src/faucet.rs]: Switched to tracing::info

File: crates/sui-graphql-client/src/faucet.rs (hunk 3 above)
Comment (stefan-mysten) [reply crates/sui-graphql-client/src/faucet.rs]: Fixed.

File: crates/sui-graphql-client/src/faucet.rs (hunk 2 above)
Comment (stefan-mysten) [reply crates/sui-graphql-client/src/faucet.rs]: That makes sense. This should probably be a separate crate, as you do not need a client to do a faucet request.

File: crates/sui-graphql-client/src/faucet.rs (hunk 1 above)
Comment (stefan-mysten) [reply crates/sui-graphql-client/src/faucet.rs]: Ok, now I understood what you were referring to. Fixed.

**Comment by stefan-mysten** [comment]
Note we'd need to probably fix the sui-sdk-types import for `sui-graphql-client` to match the crate name, and then fix the code and docs.

File: crates/sui-graphql-client/src/faucet.rs [hunk 8]
```diff
@@ -0,0 +1,236 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct BatchFaucetResponse {
+    // This string is the Uuid for the req
+    pub task: Option<String>,
+    pub error: Option<String>,
+}
+
+#[derive(serde::Deserialize)]
+pub struct FaucetResponse {
```
bmwill, line 34: This doesn't need to be pub

crates/sui-graphql-client/src/faucet.rs
```diff
@@ -0,0 +1,236 @@
...
+const FAUCET_POLL_INTERVAL: Duration = Duration::from_secs(2);
+
+pub struct FaucetClient {
+    faucet_url: Url,
+    inner: reqwest::Client,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct BatchFaucetResponse {
```
bmwill, line 27: This seems to be unused, can we remove it?

bmwill:
A few nits and a rebase then we can get this landed

stefan-mysten:
> A few nits and a rebase then we can get this landed

Thanks, I think it should be ready now (if the CI passes!)

PR #12, opened 2024-09-24
Add coins function

stefan-mysten:
I am not sure how to deal with the lifetime issues & borrow checker. `Coin` from `types` has a `'a` lifetime, and requires to pass a borrowed `Object`. I cannot figure out how to make it happy when converting from `Object` to a `Coin` such that I can return the page of coins.


bmwill:
what is the shape of the data that is being fetched from graphql? If we do want it to eventually return it as the framework "Coin" type we can change the type to use a `Cow` so that we can use an owned version (and consequently a 'static lifetime)

stefan-mysten:
> what is the shape of the data that is being fetched from graphql?

So, I am using the `objects` top-level query and filtering by owner address and type ("0x2::coin::Coin"). The `objects` function returns a list of Objects, which are converted into `Object` from the `bcs` returned by GraphQL.

stefan-mysten:
If the question was about GraphQL's Coin type, then this is more complex.
https://github.com/MystenLabs/sui-rust-sdk/blob/gql_coin_api/crates/sui-graphql-client/schema/graphql_rpc.graphql#L497-L637

bmwill:
> > what is the shape of the data that is being fetched from graphql?
> 
> So, I am using the `objects` top-level query and filtering by owner address and type (`0x2::coin::Coin`). The `objects` function returns a list of `Object`s; converting each bcs into an `Object` object. Then each object is converted into a `Coin` object through the `Coin::try_from_object`.

Ah then we should be able to use a cow with a static lifetime to solve this issue

stefan-mysten:
Nice, thanks for the suggestion. Pushed that change in another PR, let me know what you think.
https://github.com/MystenLabs/sui-rust-sdk/pull/15

stefan-mysten:
I also added a stream function for coins, and put it under `stream` feature. Should we gate it or just make it always available?

bmwill:
There's no reason to gate that feature i think. What you could do is change the tokio import to not bring in *everything* (ie not use full) but as it stands today with the graphql client crate, you don't really gain much by gating the streaming feature, as you're already essentially bringing in those deps elsewhere

stefan-mysten:
Thanks for the suggestion!

bmwill:
just need to resolve the conflicts then we can get this landed

stefan-mysten:
> just need to resolve the conflicts then we can get this landed

Resolved!

PR #13, opened 2024-09-24
Add balance function

stefan-mysten:


bmwill:
Does graphql have a balance api? Or do you need to iterate through the coins yourself to do the calculation?

stefan-mysten:
> Does graphql have a balance api? Or do you need to iterate through the coins yourself to do the calculation?

🤦  yes it exists. Forgot about it.

bmwill:
Awesome lets use that then!

crates/sui-graphql-client/src/lib.rs
```diff
@@ -194,6 +195,38 @@ impl Client {
...
+    // Balance API
+    // ===========================================================================
+
+    /// Get the balance of all the coins owned by address for the provided coin type.
+    /// Coin type will default to `0x2::coin::Coin<0x2::sui::SUI>` if not provided.
+    pub async fn balance(
+        &self,
+        address: Address,
+        coin_type: Option<&str>,
+    ) -> Result<Option<u128>, Error> {
```
bmwill, line 208: Is balance a u128 or a u64 in the graphql api?

stefan-mysten: It's a ... BigInt :D
```
BigInt

String representation of an arbitrary width, possibly signed integer.
```

stefan-mysten: Maybe then we just return the String and let the user decide how they want to parse it. 
In the CLI we've been using `u128` for balances though.

bmwill: i ask because theres no way to actually have a balance > a u64 today (I think). I'm fine if we use a u128 though, its much better than a string

PR #18, opened 2024-09-28
Add dynamic fields query

stefan-mysten:
Added `dynamic_field`, `dynamic_object_field` and `dynamic_fields` queries to client. They all return a type that includes the `serde_json::Value` and `Object`. Also added a `dynamic_fields_on_wrapped_objects` function that uses the `owner` to query for dynamic fields. 

My understanding from the GraphQL code is that `MoveObject`'s bcs is the more generic `Object`, which should work just fine.

Finally, the last few changes ([3be26a8b](https://github.com/MystenLabs/sui-rust-sdk/pull/18/commits/3be26a8bf8c7cbe524e09b2c1299cd5965b68b11) and [43fbd400](https://github.com/MystenLabs/sui-rust-sdk/pull/18/commits/43fbd40058f980c6b811df82f276b76ec143957b)) were informed by integrating this `sui-sdk-types` and `sui-graphql-client` as dependencies in another project. The CLI there relies heavily on dynamic fields queries, on wrapped objects.

### DIFF
--- a/crates/sui-graphql-client/src/lib.rs
+++ b/crates/sui-graphql-client/src/lib.rs
@@ -22,6 +22,11 @@ use query_types::CoinMetadataArgs;
 use query_types::CoinMetadataQuery;
 use query_types::DryRunArgs;
 use query_types::DryRunQuery;
+use query_types::DynamicFieldArgs;
+use query_types::DynamicFieldConnectionArgs;
+use query_types::DynamicFieldQuery;
+use query_types::DynamicFieldsOwnerQuery;
+use query_types::DynamicObjectFieldQuery;
 use query_types::EpochSummaryArgs;
 use query_types::EpochSummaryQuery;
 use query_types::EventFilter;
@@ -48,6 +53,8 @@ use query_types::TransactionMetadata;
 use query_types::TransactionsFilter;
 use query_types::Validator;
 
+use serde::de::DeserializeOwned;
+use serde::Serialize;
 use sui_types::types::framework::Coin;
 use sui_types::types::Address;
 use sui_types::types::CheckpointSequenceNumber;
@@ -58,6 +65,7 @@ use sui_types::types::SignedTransaction;
 use sui_types::types::Transaction;
 use sui_types::types::TransactionEffects;
 use sui_types::types::TransactionKind;
+use sui_types::types::TypeTag;
 use sui_types::types::UserSignature;
 
 use anyhow::anyhow;
@@ -79,11 +87,43 @@ const DEVNET_HOST: &str = "https://sui-devnet.mystenlabs.com/graphql";
 const LOCAL_HOST: &str = "http://localhost:9125/graphql";
 static USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"),);
 
+// ===========================================================================
+// Output Types
+// ===========================================================================
+
 #[derive(Debug)]
 pub struct DryRunResult {
     pub effects: Option<TransactionEffects>,
     pub error: Option<String>,
 }
+
+#[derive(Debug)]
+pub struct DynamicFieldName {
+    /// The type name of this dynamic field name
+    pub type_: TypeTag,
+    /// The bcs bytes of this dynamic field name
+    pub bcs: Vec<u8>,
+    /// The json representation of the dynamic field name
+    pub json: Option<serde_json::Value>,
+}
+
+type Typename = String;
+
+#[derive(Debug)]
+pub struct DynamicFieldOutput {
+    /// The name of the dynamic field
+    pub name: DynamicFieldName,
+    /// The dynamic field value typename and bcs
+    pub value: Option<(Typename, Vec<u8>)>,
+    /// The json representation of the dynamic field value object
+    pub value_as_json: Option<serde_json::Value>,
+}
+
+/// Helper struct for passing a value that has a type that implements Serialize, for the dynamic
+/// fields API.
+pub struct NameValue(Vec<u8>);
+/// Helper struct for passing a raw bcs value.
+pub struct BcsName(pub Vec<u8>);
 
 #[derive(Debug)]
 /// A page of items returned by the GraphQL server.
@@ -136,6 +176,34 @@ pub struct PaginationFilter<'a> {
     direction: Direction,
     cursor: Option<&'a str>,
     limit: Option<i32>,
+}
+
+impl<T: Serialize> From<T> for NameValue {
+    fn from(value: T) -> Self {
+        NameValue(bcs::to_bytes(&value).unwrap())
+    }
+}
+
+impl From<BcsName> for NameValue {
+    fn from(value: BcsName) -> Self {
+        NameValue(value.0)
+    }
+}
+
+impl DynamicFieldOutput {
+    pub fn deserialize<T: DeserializeOwned>(
+        &self,
+        expected_type: TypeTag,
+    ) -> Result<T, anyhow::Error> {
+        assert_eq!(
+            expected_type, self.name.type_,
+            "Expected type {}, but got {}",
+            expected_type, self.name.type_
+        );
+
+        let bcs = &self.name.bcs;
+        bcs::from_bytes::<T>(bcs).map_err(|_| anyhow!("Cannot decode BCS bytes"))
+    }
 }
 
 /// The GraphQL client for interacting with the Sui blockchain.
@@ -261,7 +329,7 @@ impl Client {
     /// provided.
     ///
     /// This will return `Ok(None)` if the epoch requested is not available in the GraphQL service
-    /// (e.g., due to prunning).
+    /// (e.g., due to pruning).
     pub async fn reference_gas_price(&self, epoch: Option<u64>) -> Result<Option<u64>, Error> {
         let operation = EpochSummaryQuery::build(EpochSummaryArgs { id: epoch });
         let response = self.run_query(&operation).await?;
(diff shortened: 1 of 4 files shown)


crates/sui-graphql-client/src/query_types/dynamic_fields.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// Returns the JSON representation of the field value, if available.
+    pub fn field_value_json(&self) -> Option<JsonValue> {
+        self.value.as_ref().and_then(|v| v.field_value_json())
+    }
+}
+
+impl From<DynamicField> for DynamicFieldOutput {
+    fn from(val: DynamicField) -> DynamicFieldOutput {
+        let object = match val.value {
+            Some(DynamicFieldValue::MoveObject(ref mo)) => mo
```
stefan-mysten, line 159: The BCS of a MoveObject from GraphQL is the encoded generic Object type

crates/sui-graphql-client/src/query_types/mod.rs
```diff
@@ -107,8 +117,16 @@ pub struct GQLAddress {
 #[cynic(schema = "rpc", graphql_type = "MoveObject")]
 pub struct MoveObject {
     pub bcs: Option<Base64>,
+    pub contents: Option<MoveValue>,
```
stefan-mysten, line 120: Split this up in another type for the contents field.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -478,6 +502,159 @@ impl Client {
...
+    /// Access a dynamic field on an object using its name. Names are arbitrary Move values whose
+    /// type have copy, drop, and store, and are specified using their type, and their BCS
+    /// contents, Base64 encoded.
+    ///
+    /// Same as [`dynamic_field`] but instead of passing the bcs bytes, it takes a value of a
+    /// type that implements Serialize. The `type_` parameter is still required and represents the
+    /// dynamic field's type.
+    ///
+    /// This returns [`DynamicFieldOutput`] which contains the name, the value as json, and object.
+    pub async fn dynamic_field_with_name(
```
amnn, line 518: `with_name` is a bit of a confusing name for this, because both functions accept a name, but one accepts it as a serializable type, and the other accepts BCS bytes. My assumption is that people will tend to favour this form, so maybe we name this one `dynamic_field`, and we come up with a name for the BCS version that indicates that it is more primitive? 

... `dynamic_field_with_encoded_name`, `dynamic_field_with_bcs_name?` ? (Naming is hard)

The other option is to adopt the pattern that we suggested for transaction inputs: Introduce a type for dynamic field names, and `Into` impls for that:

```rust

enum Name(Vec<u8>);
pub struct BcsName(Vec<u8>);

impl GraphQlClient {
    pub async fn dynamic_field(
        &self, 
        address: Address, 
        type_: &str, 
        name: impl Into<Name>,
    ) { /* ... */ }
}

impl<T: Serialize> From<T> for Name { /* ... */ }
impl From<BcsName> for Name { /* ... */ };
```

Now you can call it as `client.dynamic_field(addr, "...", my_type)` or `client.dynamic_field(addr, "...", BcsName(...))`, meaning you no longer need to double the API surface area.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -478,6 +502,159 @@ impl Client {
...
+    ///
+    /// This returns [`Page`] of [`DynamicFieldOutput`]s.
+    pub async fn dynamic_fields(
+        &self,
+        address: Address,
+        after: Option<String>,
+        before: Option<String>,
+        first: Option<i32>,
+        last: Option<i32>,
+    ) -> Result<Option<Page<DynamicFieldOutput>>, Error> {
```
amnn, line 626: nit: It would be great if the `Page` type itself had a way of describing an "empty" page, so we could avoid the nesting with `Option` -- it will lead people to always have to perform two checks: Is there a page? And then are there any elements in the page? If they only care about iterating through the elements of the page, then this is a bit clunky.

amnn (commented):
Couple more comments! This is great though, thanks @stefan-mysten

crates/sui-graphql-client/src/lib.rs
stefan-mysten: Added this!

crates/sui-graphql-client/src/lib.rs
stefan-mysten: See #36

crates/sui-graphql-client/src/lib.rs
```diff
@@ -145,6 +184,25 @@ impl Default for PaginationFilter<'_> {
     }
 }
 
+impl<T: Serialize> From<T> for Name {
+    fn from(value: T) -> Self {
+        Name(bcs::to_bytes(&value).unwrap())
+    }
+}
```
bmwill, line 191: I don't think this is necessarily correct given a Name includes the type and this would probably only include the value?

bmwill: or you should rename this type to indicate that its the name value and not the full name

crates/sui-graphql-client/src/lib.rs
```diff
@@ -145,6 +184,25 @@ impl Default for PaginationFilter<'_> {
...
+        Name(value.0)
+    }
+}
+
+impl DynamicFieldOutput {
+    pub fn deserialize<T: DeserializeOwned>(&self) -> Result<T, anyhow::Error> {
+        let bcs = &self.name.bcs;
+        bcs::from_bytes::<T>(bcs).map_err(|_| anyhow!("Cannot decode BCS bytes"))
+    }
+}
```
bmwill, line 204: Do you want to have a way to check or validate the type before trying to deserialize? I think that would be the ideal, safest solution is to require someone to provide the type they're expecting and have the checked before trying to deserialize the value

crates/sui-graphql-client/src/lib.rs
stefan-mysten: Ah yes, good catch. Thank you

crates/sui-graphql-client/src/lib.rs
stefan-mysten: That sounds like a good extra check, fixed.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -145,6 +184,34 @@ impl Default for PaginationFilter<'_> {
...
+    fn from(value: BcsName) -> Self {
+        NameValue(value.0)
+    }
+}
+
+impl DynamicFieldOutput {
+    pub fn deserialize<T: DeserializeOwned>(
+        &self,
+        expected_type: &str,
+    ) -> Result<T, anyhow::Error> {
```
amnn, line 203: This is a broader question than for this PR, but what are our thoughts on using `anyhow` to carry SDK errors? It gives us a lot of flexibility, but it makes the error types opaque for users (e.g. suppose I want to treat an "object not found" error differently from other errors, that's difficult).

cc @bmwill

crates/sui-graphql-client/src/lib.rs
```diff
@@ -539,6 +606,139 @@ impl Client {
...
+        &self,
+        address: Address,
+        type_: &str,
+        name: impl Into<NameValue>,
+    ) -> Result<Option<DynamicFieldOutput>, Error> {
+        let bcs = name.into().0;
+        let operation = DynamicFieldQuery::build(DynamicFieldArgs {
+            address,
+            name: crate::query_types::DynamicFieldName {
+                type_: type_.to_string(),
```
amnn, line 643: nit: if the body of this function ultimately needs to take ownership of the `type_`, then IMO it's preferable to indicate that in the type signature by accepting `type_: String`, otherwise in some scenarios you are forcing an unnecessary copy.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -539,6 +606,139 @@ impl Client {
...
+                    .collect::<Vec<DynamicFieldOutput>>();
+
+                Ok(Some(Page::new(page_info, jsons)))
+            } else {
+                Ok(None)
+            }
+        } else {
+            Ok(None)
+        }
+    }
```
amnn, line 740: `let ... else` + nested patterns:

```suggestion
        let Some(Query { owner: Some(Owner { dynamic_fields }) }) = response.data else {
            return Ok(Page::empty());
        };
        
        Ok(Page::new(
          dynamic_fields.page_info,
          dynamic_fields
              .nodes
              .into_iter()
              .map(Into::into)
              .collect()))
    }
```

crates/sui-graphql-client/src/lib.rs
```diff
@@ -765,6 +965,63 @@ impl Client {
         }
     }
 
+    /// Return the contents' JSON of an object that is a Move object.
+    ///
+    /// If the object does not exist (e.g., due to prunning), this will return `Ok(None)`.
```
amnn, line 970:
```suggestion
    /// If the object does not exist (e.g., due to pruning), this will return `Ok(None)`.
```

Also -- what about if it was deleted or wrapped?

crates/sui-graphql-client/src/lib.rs
```diff
@@ -765,6 +965,63 @@ impl Client {
...
+    /// Similarly, if this is not an object but an address, it will return `Ok(None)`.
+    pub async fn object_move_contents(
```
amnn, line 972: nit: I would probably call this `move_object_contents`, seeing as a `MoveObject` is a concept we have.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -765,6 +965,63 @@ impl Client {
...
+                .and_then(|o| o.as_move_object)
+                .and_then(|o| o.contents)
+                .and_then(|mv| mv.json))
+        } else {
+            Ok(None)
+        }
+    }
+    /// Return the BCS of an object that is a Move object.
+    ///
+    /// If the object does not exist (e.g., due to prunning), this will return `Ok(None)`.
```
amnn, line 997:
```suggestion
    /// If the object does not exist (e.g., due to pruning), this will return `Ok(None)`.
```

crates/sui-graphql-client/src/query_types/mod.rs
```diff
@@ -111,6 +120,19 @@ pub struct MoveObject {
...
 
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(schema = "rpc", graphql_type = "MoveObject")]
+pub struct MoveObjectContents {
+    pub contents: Option<MoveValue>,
+}
+
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(schema = "rpc", graphql_type = "MoveValue")]
+pub struct MoveValue {
```
amnn, line 131: I'm not sure it makes sense for the shared `MoveValue` to have both the `BCS` and `JSON` fields -- it seems unlikely that you would query for both in one query, so this would imply some over-fetching.

amnn (commented):
The main dynamic field changes LGTM, but there are some changes related to fetching objects and their contents that I think may need some work along the lines of what we talked about a week or so ago:

- That we would ideally support three or four representations of objects: JSON, BCS (raw bytes), Serializable types, and a kind of dynamic `MoveValue` (similar to JSON, but native to Move, similar to `annotated::MoveValue` in the Move world).
- This implies we would need some way to specify which kind of representation we want when querying for things that return move values (dynamic fields included).

I wonder if there's a way we can address that separately though (in another PR) because it seems a little out of scope of this PR, even though it has implications for what's going on here.

crates/sui-graphql-client/src/lib.rs
bmwill: Ideally we don't use anyhow at all and instead have more detailed errors. anyhow doesn't impl Error itself so it doesn't really play very nicely when used in a library. On the far other end of the spectrum I'm very against the SuiError type of thing where we have a single bucket where there's a variant for everything. Errors should try to be focused and when they need to be more general we can utilize the "source" paradigm where we curry a Box<dyn Error> inside of our error that users can downcast to if they so desire.

crates/sui-graphql-client/src/lib.rs
bmwill: Do we want to use the TypeTag type directly instead of Strings?

bmwill:
> The main dynamic field changes LGTM, but there are some changes related to fetching objects and their contents that I think may need some work along the lines of what we talked about a week or so ago:
> 
> * That we would ideally support three or four representations of objects: JSON, BCS (raw bytes), Serializable types, and a kind of dynamic `MoveValue` (similar to JSON, but native to Move, similar to `annotated::MoveValue` in the Move world).
> * This implies we would need some way to specify which kind of representation we want when querying for things that return move values (dynamic fields included).
> 
> I wonder if there's a way we can address that separately though (in another PR) because it seems a little out of scope of this PR, even though it has implications for what's going on here.

One approach to this would be to tweak the API to instead of having a single function that takes all the parameters, instead you have a request builder paradigm (like reqwest) where you build up your request and specify the "format" you want. The question here though, is, do you ever want multiple types of formats at the same type. Also how much do we want to bake into this "builtin" client vs pushing folks to craft their own queries

crates/sui-graphql-client/src/lib.rs
stefan-mysten: That's next on my agenda, to rework all the anyhow errors into explicit errors.

crates/sui-graphql-client/src/query_types/mod.rs
stefan-mysten: The `DynamicFieldOutput` has a `json` field, which returns the value as a JSON, and the `value` field which returns the BCS and typename. The thought here was that if you know/have the type you could reconstruct it from BCS, but if you only need a field and don't have the type, you use the JSON object.

crates/sui-graphql-client/src/lib.rs
stefan-mysten: Fixed

amnn:
Builder API is intriguing -- I think it's fairly unlikely that someone would want multiple representations of a type (unlikely enough that we don't need to support it in the built-in client), but I think the SDK does need to have some native notion of a "Move Value", because in the GraphQL schema these are represented as scalars, which means it's not really a question of writing a custom query -- these types exist in the leaves of the query, so even if someone wrote a custom query, they would need some help to translate the JSON blob at that leaf into a Rust representation of a Move value and that is a universal problem.

PR #21, opened 2024-09-30
Add active validators

stefan-mysten:


crates/sui-graphql-client/src/query_types/mod.rs
```diff
@@ -92,17 +113,17 @@ impl TryFrom<BigInt> for u64 {
     }
 }
 
-impl From<Address> for SuiAddress {
-    fn from(value: Address) -> Self {
+impl From<NativeAddress> for SuiAddress {
+    fn from(value: NativeAddress) -> Self {
         SuiAddress(value.to_string())
     }
 }
```
bmwill, line 120: Is there a way to just have SuiAddress be defined as Address?

bmwill: seems like maybe https://cynic-rs.dev/derives/scalars.html#impl_scalar?

stefan-mysten: Thanks for the suggestion. This would be to reduce confusion / have fewer types one has to reason with?

bmwill: yes correct, and improve ergonomics

stefan-mysten: Ready to do it in a follow up PR once this is merged(to not bring a lot of non-relevant changes here).

PR #23, opened 2024-09-30
Add dry run

stefan-mysten:
Note: will need to add tests later once we fix the transaction bcs issue (intent is part of bcs).

### DIFF
--- a/crates/sui-graphql-client/src/lib.rs
+++ b/crates/sui-graphql-client/src/lib.rs
@@ -747,23 +747,18 @@ impl Client {
             .as_ref()
             .and_then(|tx| tx.dry_run_transaction_block.error.clone());
 
-        // This seems to fail as it is not able to decode the transaction effects' bcs from GraphQL
-        // into sui-types TransactionEffects type.
         let effects = response
             .data
             .map(|tx| tx.dry_run_transaction_block)
-            .and_then(|tx| {
-                tx.transaction
-                    .and_then(|tx| tx.effects)
-                    .and_then(|e| e.bcs)
-                    .map(|bcs| bcs::from_bytes::<TransactionEffects>(bcs.0.as_ref()))
-            })
+            .and_then(|tx| tx.transaction)
+            .and_then(|tx| tx.effects)
+            .and_then(|bcs| bcs.bcs)
+            .map(|bcs| base64ct::Base64::decode_vec(bcs.0.as_str()))
             .transpose()
-            .map_err(|e| {
-                Error::msg(format!(
-                    "Cannot decode bcs bytes into TransactionEffects. {e}"
-                ))
-            })?;
+            .map_err(|_| Error::msg("Cannot decode bcs bytes from Base64 for transaction effects"))?
+            .map(|bcs| bcs::from_bytes::<TransactionEffects>(&bcs))
+            .transpose()
+            .map_err(|_| Error::msg("Cannot decode bcs bytes into TransactionEffects"))?;
 
         Ok(DryRunResult { effects, error })
     }
@@ -1160,8 +1155,6 @@ mod tests {
     }
 
     #[tokio::test]
-    #[ignore]
-    // TODO this fails because it cannot decode the transaction effects properly.
     async fn test_dry_run() {
         let client = Client::new_testnet();
         // this tx bytes works on testnet
--- a/crates/sui-graphql-client/src/lib.rs
+++ b/crates/sui-graphql-client/src/lib.rs
@@ -20,6 +20,7 @@ use query_types::CoinMetadataArgs;
 use query_types::CoinMetadataQuery;
 use query_types::DryRunArgs;
 use query_types::DryRunQuery;
+use query_types::DryRunResult;
 use query_types::EpochSummaryArgs;
 use query_types::EpochSummaryQuery;
 use query_types::EventFilter;
@@ -689,7 +690,7 @@ impl Client {
         &self,
         tx: &Transaction,
         skip_checks: Option<bool>,
-    ) -> Result<Option<Transaction>, Error> {
+    ) -> Result<Option<DryRunResult>, Error> {
         let tx_bytes = base64ct::Base64::encode_string(
             &bcs::to_bytes(&tx).map_err(|_| Error::msg("Cannot encode Transaction as BCS"))?,
         );
@@ -708,7 +709,7 @@ impl Client {
         tx_kind: &TransactionKind,
         skip_checks: Option<bool>,
         tx_meta: TransactionMetadata,
-    ) -> Result<Option<Transaction>, Error> {
+    ) -> Result<Option<DryRunResult>, Error> {
         let tx_bytes = base64ct::Base64::encode_string(
             &bcs::to_bytes(&tx_kind).map_err(|_| Error::msg("Cannot encode Transaction as BCS"))?,
         );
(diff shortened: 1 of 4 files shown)


crates/sui-graphql-client/src/lib.rs
```diff
@@ -602,6 +602,55 @@ impl Client {
...
+    ///
+    ///  - if present, txBytes is assumed to be a [`TransactionKind`],
+    ///  - if absent, txBytes is assumed to be [`Transaction`].
+    ///
+    /// `skipChecks` optional flag to disable the usual verification checks that prevent access to
+    /// objects that are owned by addresses other than the sender, and calling non-public,
+    /// non-entry functions, and some other checks. Defaults to false.
+    pub async fn dry_run(
+        &self,
+        tx_bytes: Vec<u8>,
```
bmwill, line 619: Same comment as in the other PR, this should be a &Transaction not bytes

stefan-mysten: Fixed!

crates/sui-graphql-client/src/lib.rs
```diff
@@ -602,6 +602,57 @@ impl Client {
...
+    /// `txBytes` is either a TransactionData struct or a [`TransactionKind`] struct, BCS-encoded
+    /// and then Base64-encoded. The expected type is controlled by the presence or absence of
+    /// `txMeta`:
+    ///
+    ///  - if present, txBytes is assumed to be a [`TransactionKind`],
+    ///  - if absent, txBytes is assumed to be [`Transaction`].
+    ///
+    /// `skipChecks` optional flag to disable the usual verification checks that prevent access to
+    /// objects that are owned by addresses other than the sender, and calling non-public,
+    /// non-entry functions, and some other checks. Defaults to false.
```
bmwill, line 616: Do we need to have two dry_run apis? This doc comment doesn't match the function signature now that it takes &Transaction...

stefan-mysten: I don't follow the `two dry_run apis`? Were you referring to the usage of `txMeta` to determine the type of the transaction to dry run?

I assume we can always just use `Transaction`, and do not pass the `txMeta` argument. I am not sure what was the rationale behind having this two-version dry-run in GraphQL.

bmwill: your comment says that the metadata only is used if a TransactionKind is sent, so a good api would split these two cases into separate apis, a Kind with metadata and a Transaction only api

stefan-mysten: Ah I see what you mean. Thanks for that!

stefan-mysten: Refactored

crates/sui-graphql-client/src/lib.rs
```diff
@@ -683,6 +688,82 @@ impl Client {
...
+    /// Dry run a [`Transaction`] and return the transaction.
+    ///
+    /// `skipChecks` optional flag disables the usual verification checks that prevent access to
+    /// objects that are owned by addresses other than the sender, and calling non-public,
+    /// non-entry functions, and some other checks. Defaults to false.
+    pub async fn dry_run_tx(
+        &self,
+        tx: &Transaction,
+        skip_checks: Option<bool>,
+    ) -> Result<Option<Transaction>, Error> {
```
bmwill, line 704: This return type doesn't make any sense. If i'm performing a dryrun i want to preview the effects of the transaction, not get back the exact txn that i sent

crates/sui-graphql-client/src/lib.rs
```diff
@@ -683,6 +688,82 @@ impl Client {
...
+    /// objects that are owned by addresses other than the sender, and calling non-public,
+    /// non-entry functions, and some other checks. Defaults to false.
+    ///
+    /// `tx_meta` is the transaction metadata.
+    pub async fn dry_run_tx_kind(
+        &self,
+        tx_kind: &TransactionKind,
+        skip_checks: Option<bool>,
+        tx_meta: TransactionMetadata,
+    ) -> Result<Option<Transaction>, Error> {
```
bmwill, line 723: Same here

crates/sui-graphql-client/src/lib.rs
stefan-mysten: 🤦  yes, you're right.

crates/sui-graphql-client/src/query_types/dry_run.rs
```diff
@@ -0,0 +1,120 @@
...
+pub struct Input {
+    pub __typename: String,
+    pub ix: i32,
+}
+
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(schema = "rpc", graphql_type = "GasCoin")]
+pub struct GasCoin {
+    #[cynic(rename = "_")]
+    pub __underscore: Option<bool>,
```
stefan-mysten, line 36: The GraphQL schema has a "dummy" field.

crates/sui-graphql-client/src/query_types/dry_run.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(schema = "rpc", graphql_type = "DryRunResult")]
+pub struct DryRunResult {
+    pub error: Option<String>,
+    pub results: Option<Vec<DryRunEffect>>,
+}
```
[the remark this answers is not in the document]
stefan-mysten: TIL: Returning the transaction from the dry-run gives us the effects as well, so we could just build the `TransactionEffects` from bcs and return only the effects. I will do that!

crates/sui-graphql-client/src/query_types/dry_run.rs
```diff
@@ -0,0 +1,59 @@
...
+
+#[derive(cynic::InputObject, Debug)]
+#[cynic(schema = "rpc", graphql_type = "ObjectRef")]
+pub struct ObjectRef {
+    pub address: Address,
+    pub digest: String,
+    pub version: Uint53,
+}
+
+impl From<ObjectReference> for ObjectRef {
```
stefan-mysten, line 50: This is useful when passing the required transaction metadata object, which includes the gas objects that is a list of `ObjectRef`, so one can just do `.into()` on those `ObjectReference`.

crates/sui-graphql-client/src/query_types/dry_run.rs
```diff
@@ -0,0 +1,58 @@
...
+pub struct DryRunQuery {
+    #[arguments(txBytes: $tx_bytes, skipChecks: $skip_checks, txMeta: $tx_meta)]
+    pub dry_run_transaction_block: DryRunResult,
+}
+
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(schema = "rpc", graphql_type = "DryRunResult")]
+pub struct DryRunResult {
+    pub error: Option<String>,
+    pub transaction: Option<TransactionBlock>,
```
bmwill, line 21: What is a TransactionBlock? the definition here only has a single bcs field which doesn't indicate if it includes events, effects, ect

stefan-mysten: Ah yes,  thanks for catching this mistake.

GraphQL calls TransactionBlock the type that includes most data (gas, effects, signatures, kind, sender, digest, expiration). GraphQL dry run will return the transaction block with its effects. 
The bcs of TransactionBlock is the `SenderSignedData`, bcs serialized and Base64 encoded, so it does not contain the effects. I'd need to add the `effects` field on the `TransactionBlock` and use the bcs field to convert to `sui-sdk-types::types::TransactionEffects`.

stefan-mysten: @bmwill fixed it now.

PR #24, opened 2024-10-01
Add more stream functions

stefan-mysten:


crates/sui-graphql-client/src/lib.rs
```diff
@@ -322,8 +323,8 @@ impl Client {
     pub async fn coins(
         &self,
         owner: Address,
-        after: Option<&str>,
-        before: Option<&str>,
+        after: Option<String>,
+        before: Option<String>,
```
bmwill, line 327: why require an owned string now?

stefan-mysten: I think I did like this because I need to pass the event filter in the streams calls, so I needed to clone it. Maybe I can just pass a reference.

stefan-mysten: @bmwill this is what I came up with, what do you think?

bmwill:
@stefan-mysten I wonder...this PR has quite a bit of repetitive code, would it be possible to generalize this and have a generic "stream" query function that can be reused everywhere? As in we have a function for issuing a generic graphql request, can we have something similar (with the constraint that we may have paging information that is standardized) that will be used as inputs into the query, we could then let anyone trivially run a pagable query as a stream

stefan-mysten:
@bmwill Interesting idea, yes, I can think we can try that. I'll give it a go and see if I can make it work!

stefan-mysten:
I looked into this over the weekend. I think it's doable, the only issue is how to "rebuild" the query with a new set of variables (because the cursor changes and we need to re-run the query to fetch the next page with a different after) programmatically. I stumbled a bit over this and couldn't find a clean and straightforward way.

stefan-mysten:
Superseded by #51.

PR #25, opened 2024-10-01
sui-crypto: introduce multisig verifier and aggregator

bmwill:


jnaulty (approved):
thank you for adding multisig + zkLogin support to the rust SDK!

PR #29, opened 2024-10-03
Add impl_scalar for u64 and Uint53

stefan-mysten:


crates/sui-graphql-client/README.md
```diff
@@ -132,19 +133,15 @@ pub struct CustomQuery {
 
 #[derive(cynic::QueryFragment, Debug)]
 pub struct Epoch {
-    pub epoch_id: Uint53,
+    pub epoch_id: u64,
     pub reference_gas_price: Option<BigInt>,
```
bmwill, line 137: can we do the same with BigInt?

stefan-mysten: I will try in a separate PR.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -989,6 +977,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[ignore]
```
stefan-mysten, line 980: It's been failing repeatedly due to timeouts, so ignoring for now.

PR #30, opened 2024-10-07
Add package query

stefan-mysten:
This is needed in one of the internal tools that we're working on.


crates/sui-graphql-client/src/lib.rs
```diff
@@ -671,6 +673,44 @@ impl Client {
...
+
+    // TODO: We should return maybe the MovePackage type, but the `data` field is not accessible.
+    // We might want to expose an API for getting the package or the struct depending what the
+    // object is.
+    /// Get a package's content by its address and optionally, a specific version.
+    pub async fn package(
+        &self,
+        address: Address,
+        version: Option<u64>,
+    ) -> Result<Option<Object>, Error> {
```
bmwill, line 688: What value does this have over just a normal object lookup since its just returning an Object anyway?

stefan-mysten: I was hoping we can return a move package instead, but it seems that the BCS returned by GraphQL is the object. So indeed, no need for it.

bmwill: i would think there would be a way to have graphql return the normalized package

PR #32, opened 2024-10-11
Fix transaction(s) query

stefan-mysten:
The `TransactionBlock` bcs in GraphQL now encodes the `TransactionData`. This PR fixes the logic to correctly deserialize the transaction bcs data and signatures into a `SignedTransaction`. 

As the service does not serve yet the new data, I tested it locally by starting a local network and GraphQL server. For now, I put an `#ignore` on the test.

### DIFF
--- a/crates/sui-graphql-client/src/lib.rs
+++ b/crates/sui-graphql-client/src/lib.rs
@@ -775,12 +775,12 @@ impl Client {
         });
         let response = self.run_query(&operation).await?;
 
-        Ok(response
+        response
             .data
             .and_then(|d| d.transaction_block)
             .map(|tx| tx.try_into())
             .transpose()
-            .map_err(|e| Error::msg(format!("Cannot decode transaction: {e}")))?)
+            .map_err(|e| Error::msg(format!("Cannot decode transaction: {e}")))
     }
 
     /// Get a page of transactions based on the provided filters.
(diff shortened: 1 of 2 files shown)


crates/sui-graphql-client/src/lib.rs
```diff
@@ -809,28 +805,12 @@ impl Client {
...
 
-            let transactions = bcs
-                .iter()
-                .map(|tx| bcs::from_bytes::<SignedTransaction>(tx))
-                .collect::<Result<Vec<_>, bcs::Error>>()
-                .map_err(|e| {
-                    Error::msg(format!(
-                        "Cannot decode bcs bytes into SignedTransaction: {e}"
-                    ))
-                })?;
```
stefan-mysten, line 833: I moved this logic and fixed it into a TryFrom

crates/sui-graphql-client/src/lib.rs
```diff
@@ -775,26 +775,22 @@ impl Client {
...
-            .and_then(|tbq| tbq.transaction_block)
-            .and_then(|tb| tb.bcs)
-            .map(|bcs| base64ct::Base64::decode_vec(bcs.0.as_str()))
+            .and_then(|d| d.transaction_block)
+            .map(|tx| tx.try_into())
             .transpose()
-            .map_err(|e| Error::msg(format!("Cannot decode Base64 transaction bcs bytes: {e}")))?
-            .map(|bcs| bcs::from_bytes::<SignedTransaction>(&bcs))
-            .transpose()?;
-        Ok(signed_tx)
```
stefan-mysten, line 787: I moved this logic and fixed it into a TryFrom

PR #33, opened 2024-10-11
[GraphQL Client] Add package queries

stefan-mysten:
Not sure if we want to expose the `MovePackage` or not, but this is what I came up with, which replicates what exists on the GraphQL side. 
As we discussed, having more queries that return the normalized package data would be useful.

Note that this requires the new schema in PR #31 and a bump to use Sui `v1.36.1` binary.


stefan-mysten:
Note that tests fail because the new schema has not made it yet into the `mainnet` service.

.github/workflows/ci.yml
```diff
@@ -63,7 +63,6 @@ jobs:
       - name: Run tests in wasm
         run: make wasm
 
-
```
stefan-mysten, line 66: Ha, I think I set up an autoformatter for yml/yaml files, so things got fixed.

PR #34, opened 2024-10-11
Add checkpoints query

stefan-mysten:


crates/sui-graphql-client/src/lib.rs
```diff
@@ -458,6 +460,42 @@ impl Client {
             .ok_or_else(|| Error::msg("No data in response"))?
     }
 
+    /// Get a page of [`CheckpointSummary`] for the provided parameters.
+    pub async fn checkpoints(
+        &self,
+        after: Option<&str>,
+        before: Option<&str>,
+        first: Option<i32>,
+        last: Option<i32>,
```
bmwill, line 469: maybe we should have people pass the Args themselves vs all these optional fields?

stefan-mysten: I will rework this into a pagination kind of type, with forward/backward, cursor, limit.

PR #36, opened 2024-10-18
[GraphQL Client] Rework the Page type and APIs that return Option<Page> 

stefan-mysten:
This PR simplifies the return type of several APIs from `Result<Option<Page>, Error>` to `Result<Page, Error>`. 
The Page type has now a `is_empty` function to check if there is no information returned in this page.

It should simplify a bit working with the results of these APIs.

### DIFF
--- a/crates/sui-graphql-client/src/lib.rs
+++ b/crates/sui-graphql-client/src/lib.rs
@@ -407,7 +407,7 @@ impl Client {
 
                 if !response.is_empty() {
                     for object in response.data() {
-                        if let Some(coin) = Coin::try_from_object(&object) {
+                        if let Some(coin) = Coin::try_from_object(object) {
                             yield coin.into_owned();
                         }
                     }
--- a/crates/sui-graphql-client/src/lib.rs
+++ b/crates/sui-graphql-client/src/lib.rs
@@ -93,16 +93,29 @@ pub struct Page<T> {
 }
 
 impl<T> Page<T> {
+    /// Return the page information.
     pub fn page_info(&self) -> &PageInfo {
         &self.page_info
     }
 
+    /// Return the data in the page.
     pub fn data(&self) -> &[T] {
         &self.data
     }
 
+    /// Internal function to create a new page with the provided data and page information.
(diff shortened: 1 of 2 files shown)


crates/sui-graphql-client/src/lib.rs
```diff
@@ -393,14 +405,14 @@ impl Client {
                     None,
                 ).await?;
 
-                if let Some(page) = response {
-                    for object in page.data {
-                        if let Some(coin) = Coin::try_from_object(&object) {
+                if !response.is_empty() {
```
amnn, line 408: the point of the simplification is that you no longer need this outer check. This code can be written as:

```rust
for object in response.data() {
    if let Some(coin) = Coin::try_from_object(object) {
      yield coin.into_owned();
    }
}

after = response.page_info.end_cursor;
if after.is_none() {
  break;
}
```

crates/sui-graphql-client/src/lib.rs
```diff
@@ -93,17 +93,30 @@ pub struct Page<T> {
...
+    /// Check if the page has data.
+    pub fn is_empty(&self) -> bool {
+        self.data.is_empty()
+    }
+
+    /// Internal function to create a page with no data.
+    fn new_empty() -> Self {
+        Self::new(PageInfo::default(), vec![])
+    }
 }
```
amnn, line 120: You could also `derive(Default)` on `Page` for this.

amnn (commented):
Some small notes!

crates/sui-graphql-client/src/lib.rs
stefan-mysten: Thanks! I remember trying this and it was complaining about something :D

amnn: Ah that's a shame, thanks for checking. The issue is that `derive(Default)` introduces a `Default` constraint on all type parameters, even if it's not strictly necessary (e.g. the `Default` impl for `Vec<T>` does not require `Default` on the `T`).

PR #37, opened 2024-10-18
[GraphQL Client] Add a pagination filter

stefan-mysten:
This PR adds a pagination filter instead of explicitly requiring after/before/first/last, which was quite wrong. A user cannot pass a cursor randomly, it needs to come from a page information that was returned by GraphQL.

### DIFF
--- a/crates/sui-graphql-client/src/lib.rs
+++ b/crates/sui-graphql-client/src/lib.rs
@@ -121,6 +121,30 @@ impl<T> Page<T> {
     }
 }
 
+/// Pagination direction.
+pub enum Direction {
+    Forward,
+    Backward,
+}
+
+/// Pagination options for querying the GraphQL server. It defaults to forward pagination with a
+/// limit of 10 items per request.
+pub struct PaginationFilter<'a> {
+    direction: Direction,
+    cursor: Option<&'a str>,
+    limit: Option<i32>,
+}
+
+impl Default for PaginationFilter<'_> {
+    fn default() -> Self {
+        Self {
+            direction: Direction::Forward,
+            cursor: None,
+            limit: Some(10),
+        }
+    }
+}
+
 /// The GraphQL client for interacting with the Sui blockchain.
 /// By default, it uses the `reqwest` crate as the HTTP client.
 pub struct Client {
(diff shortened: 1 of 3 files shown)


crates/sui-graphql-client/src/lib.rs
```diff
@@ -121,6 +121,30 @@ impl<T> Page<T> {
...
+impl Default for PaginationFilter<'_> {
+    fn default() -> Self {
+        Self {
+            direction: Direction::Forward,
+            cursor: None,
+            limit: None,
+        }
+    }
+}
+
```
amnn, line 147: This should be a simpler way to achieve the same thing:

```suggestion
/// Pagination direction.
#[derive(Default)]
pub enum Direction {
    #[default]
    Forward,
    Backward,
}

/// Pagination options for querying the GraphQL server. It defaults to forward pagination with the
/// GraphQL server's default items per page limit.
#[derive(Default)
pub struct PaginationFilter<'a> {
    direction: Direction,
    cursor: Option<&'a str>,
    limit: Option<i32>,
}
```

crates/sui-graphql-client/src/lib.rs
```diff
@@ -262,18 +286,16 @@ impl Client {
...
     /// If no epoch is provided, it will return the active validators for the current epoch.
-    pub async fn active_validators(
+    pub async fn active_validators<'a>(
         &self,
         epoch: Option<u64>,
-        after: Option<String>,
-        before: Option<String>,
-        first: Option<i32>,
-        last: Option<i32>,
+        pagination_filter: Option<PaginationFilter<'a>>,
```
amnn, line 292: Complementary to the question about `Option<Page<...>>` in the output, do we need the `Option` here? Seems like it would be just as easy for someone to pass `PaginationFilter::default()` as `None`, and it makes it clear what the parameter is.

amnn (commented):
(Again, nits only!)

crates/sui-graphql-client/src/lib.rs
stefan-mysten: Thanks!

crates/sui-graphql-client/src/lib.rs
stefan-mysten: Sure, we can add that.

PR #38, opened 2024-10-18
[CI+GraphQL] Download Sui binary for localnet running in the CI + have tests only run on localnet 

stefan-mysten:
This updates the CI to enable a local network to be run from a testnet sui binary. It also splits the makefile tests into two: one for those tests that do not require a network, and one for those that need, e.g., graphql-client and soon sui-transaction-builder.

Changed tests to run on a local network. That means that things can still break if the schema is updated without updating to the binary that contains the released schema, so if that happens often, we might want to add a check on that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,21 +15,6 @@ env:
 jobs:
   test:
     runs-on: ubuntu-latest
-    services:
-      postgres: # we need this postgres instance for running a local network with indexer and graphql
-        image: postgres
-        env:
-          POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: postgrespw
-          POSTGRES_DB: sui_indexer_v2
-          POSTGRES_HOST_AUTH_METHOD: trust
-        options: >-
-          --health-cmd pg_isready
-          --health-interval 10s
-          --health-timeout 5s
-          --health-retries 5
-        ports:
-          - 5432:5432
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -15,6 +15,21 @@ env:
...
+          POSTGRES_PASSWORD: postgrespw
+          POSTGRES_DB: sui_indexer_v2
+          POSTGRES_HOST_AUTH_METHOD: trust
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
```
bmwill, line 32: Can we move/factor this out into a separate job that is specific for network tests instead of having it a part of this job?

crates/sui-graphql-client/src/lib.rs
```diff
@@ -965,239 +964,205 @@ mod tests {
...
-            let client = Client::new(n).unwrap();
-            let av = client.active_validators(None, None).await;
-            assert!(
-                av.is_ok(),
-                "Active validators query failed for network: {n}. Error: {}",
-                av.unwrap_err()
-            );
+        // this test requires to have at least epoch that finished, so wait for 10s just to be on
+        // the safe side
+        tokio::time::sleep(tokio::time::Duration::from_secs(10)).await;
```
bmwill, line 1019: Can we do this sleep outside of the test and instead on the startup of the local network?

crates/sui-graphql-client/src/lib.rs
```diff
@@ -965,239 +964,205 @@ mod tests {
 
     #[tokio::test]
     async fn test_balance_query() {
-        for (n, _) in NETWORKS.iter() {
-            let client = Client::new(n).unwrap();
-            let balance = client.balance("0x1".parse().unwrap(), None).await;
-            assert!(balance.is_ok(), "Balance query failed for network: {n}");
-        }
+        let client = Client::new_localhost();
```
bmwill, line 967: Can we maybe have a "get_test_client" method that is cfg(test) that reads from an Env Var for the url to use? This would give us the flexibility to run all of these network tests on every network we want (although we only care about doing it for this local env today)

.github/workflows/ci.yml
stefan-mysten: Yes, sounds good!

crates/sui-graphql-client/src/lib.rs
stefan-mysten: Oh, yes, that sounds like a better idea!

crates/sui-graphql-client/src/lib.rs
stefan-mysten: Yes, that sounds like a good idea! Thanks for the suggestion.

.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: clippy
         run: make clippy
 
-      - name: Run tests
+      - name: Run tests that do not require local network
```
bmwill, line 40:
```suggestion
      - name: Run tests
```

.github/workflows/ci.yml
```diff
@@ -62,3 +62,58 @@ jobs:
 
       - name: Run tests in wasm
         run: make wasm
+
+
+  run_tests_on_network:
```
bmwill, line 67:
```suggestion
  run_tests_with_network:
```

.github/workflows/ci.yml
```diff
@@ -62,3 +62,58 @@ jobs:
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: rust version
+        run: |
+          rustc --version
+          cargo --version
+      
+      - uses: taiki-e/install-action@cargo-hack
```
bmwill, line 95:
```suggestion
```

PR #39, opened 2024-10-21
[GraphQL Client[ Address nits and use de-facto Default derive

stefan-mysten:


crates/sui-graphql-client/src/lib.rs
```diff
@@ -494,20 +492,30 @@ impl Client {
...
+                pagination_filter.limit,
+                None,
+            ),
+            Direction::Backward => (
+                None,
+                pagination_filter.cursor,
+                None,
+                pagination_filter.limit,
+            ),
+        };
```
bmwill, line 512: can we factor this out into a private method on PaginationFilter do do this so that it isn't copy-pasted everywhere?

stefan-mysten: Definitely, don't know why I left it like that 🤦

stefan-mysten: Done

PR #43, opened 2024-10-23
Make pagination filter cursor to be a String instead `&str`.

stefan-mysten:
Make PaginationFilter accept a String to avoid complex liftetime issues. Also, make its fields public


crates/sui-graphql-client/src/lib.rs
```diff
@@ -172,10 +172,10 @@ pub enum Direction {
...
 /// GraphQL server's default items per page limit.
 #[derive(Default)]
-pub struct PaginationFilter<'a> {
-    direction: Direction,
-    cursor: Option<&'a str>,
-    limit: Option<i32>,
+pub struct PaginationFilter {
+    pub direction: Direction,
+    pub cursor: Option<String>,
+    pub limit: Option<i32>,
```
bmwill, line 178: Whats the rationale for changing this? We could have this be a Cow

stefan-mysten: The rationale when working with the cursor is that if it is a `&'a str`, I was having a hard time to figuring out how to make it work in a `loop`. Basically I need to do something like this

```
let mut cursor = None;

loop {
          let page = mainnet_client
            .dynamic_fields(
                Address::from_str(&app_registry_id.to_string())?,
                PaginationFilter {
                    cursor: cursor.clone(),
                    ..Default::default()
                },
            )
            .await?;
        /// do something which each dynamic field
           for dynamic_field_info in page.data().iter() {}
       if !page.page_info().has_next_page {
          break;
       }

      cursor = page.end_cursor().clone().as_deref(); --> Note that cursor is &str, so this will complain that temporary value is freed at the end of this statement.
      
}
```

stefan-mysten: So probably a `Cow` could work here.

stefan-mysten: I switched to `Cow<'a, str>` and I ended up doing this

```
        cursor = page
            .page_info()
            .end_cursor
            .as_ref()
            .map(|x| Cow::Owned(x.to_string()));

```

stefan-mysten: Don't know if that's better than a `String`. Admittedly, I still haven't gotten down this lifetime stuff very well.

stefan-mysten: Also, the `cursor` that comes from a GraphQL connection is a `String`.

PR #45, opened 2024-10-24
[GraphQL Client] Move from anyhow to custom errors

stefan-mysten:
This PR attempts to move away from anyhow and use custom errors.

Ideally, we should merge #43 #44 before this one to avoid conflicts.


stefan-mysten:
Superseded by #58

PR #47, opened 2024-10-26
[GraphQL Client] Add total transaction blocks query

stefan-mysten:
This PR adds total transaction blocks queries that return the number of tx blocks up to the specified checkpoint. 
In addition, it fixes two issues on checkpoint digests
- moves the digest to be a Digest type
- checkpoint digests were not base58 decoded

### DIFF
--- a/crates/sui-graphql-client/src/lib.rs
+++ b/crates/sui-graphql-client/src/lib.rs
@@ -59,6 +59,7 @@ use sui_types::types::framework::Coin;
 use sui_types::types::Address;
 use sui_types::types::CheckpointSequenceNumber;
 use sui_types::types::CheckpointSummary;
+use sui_types::types::Digest;
 use sui_types::types::Event;
 use sui_types::types::Object;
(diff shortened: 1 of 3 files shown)


crates/sui-graphql-client/src/lib.rs
```diff
@@ -426,6 +428,38 @@ impl Client {
...
+    /// provided, it will return the total number of transactions by the end of the last known
+    /// checkpoint.
+    pub async fn total_transaction_blocks(
+        &self,
+        digest: Option<String>,
+        seq_num: Option<u64>,
+    ) -> Result<Option<u64>, Error> {
+        ensure!(
+            !(digest.is_some() && seq_num.is_some()),
+            "Either digest or seq_num must be provided"
```
bmwill, line 442: This message conflicts with your doc comment. Its either a digest or a sequence number or neither.

Also this type of api isn't really that great as it forces you to always provide "None" to at least one of the params. Also the digest should likely be strongly typed vs stringly typed.

PR #53, opened 2024-10-31
[GraphQL Client] Initialize the max page size in a lazy way

stefan-mysten:
This fixes the previous code where the `pagination_filter` would make a request to the service config to fetch the max page size when the limit is not provided. In the case of streams, this means that every time the cursor changed, a new filter needs to be created and a request to the service config for fetching the max page size.

It now will lazily instantiate that max page size value to avoid subsequent RPC requests.


crates/sui-graphql-client/src/lib.rs
```diff
@@ -327,27 +329,42 @@ impl Client {
...
-            let cfg = self.service_config().await;
-            if let Ok(cfg) = cfg {
-                cfg.max_page_size
-            } else {
-                DEFAULT_ITEMS_PER_PAGE
-            }
-        };
+        let limit = pagination_filter
+            .limit
+            .unwrap_or(self.max_page_size().await.unwrap_or(DEFAULT_ITEMS_PER_PAGE));
```
stefan-mysten, line 339: I hate that clippy forces nowadays to use `unwrap_or`, instead of `unwrap_or_else` for lazily evaluating this statement.

crates/sui-graphql-client/src/lib.rs
```diff
@@ -115,7 +116,7 @@ const TESTNET_HOST: &str = "https://sui-testnet.mystenlabs.com/graphql";
 const DEVNET_HOST: &str = "https://sui-devnet.mystenlabs.com/graphql";
 const LOCAL_HOST: &str = "http://localhost:9125/graphql";
 static USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"),);
-
+static MAX_PAGE_SIZE: OnceCell<i32> = OnceCell::const_new();
```
bmwill, line 119: Lets instead have this information live in the Client itself. That way its specific to the endpoint and not a globally enforced value. Imagine the case where you have two clients, each connected with a separate endpoint with different max page size settings.

bmwill (approved):

crates/sui-graphql-client/src/lib.rs
stefan-mysten: Went that route initially but did not think of the case of having two clients at different endpoints! Thanks for the fix.

PR #55, opened 2024-11-01
[GraphQL Client] Add transaction effects query

stefan-mysten:
This PR adds a transaction effects query, a multi transaction effects query, and corresponding stream version.

It also splits the existing query types into two, one for `SignedTransaction` data and one for `TransactionEffects` to avoid over fetching data from the RPC.


crates/sui-graphql-client/src/lib.rs
```diff
@@ -1482,6 +1484,24 @@ impl Client {
             .map_err(|e| Error::msg(format!("Cannot decode transaction: {e}")))
     }
 
+    /// Get a transaction's effects by its digest.
+    pub async fn transaction_effects(
+        &self,
+        digest: Digest,
```
bmwill, line 1490: Lets use TransactionDigest vs Digest here to better indicate how its fetched.

PR #57, opened 2024-11-04
[GraphQL Client] Add a build crate to allow registering the schema in other crates

stefan-mysten:
Found out that registering a schema is required as `cynic` does not export it in a way that it can be trivially added to external crates that need to use `sui-graphql-client`.

This PR adds a new crate `sui-graphql-client-build` which will keep the schema file and that has a function that can be called by any crate to register the schema needed to build custom queries on top of using `sui-graphql-client`. To avoid confusion, we also use it in the `sui-graphql-client`.


crates/sui-graphql-client-build/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+/// #[tokio::main]
+/// async fn main() {
+///     let client = Client::new_mainnet();
+///     let operation = MyQuery::build(());
+///     let q = client.run_query(&operation).await.unwrap();
+///     println!("{:?}", q);
+/// }
+/// ```
+pub fn register_schema(schema_name: &str) {
+    let sdl = include_bytes!("../schema.graphql");
```
amnn, line 45: Use [`include_str`](https://doc.rust-lang.org/std/macro.include_str.html) to avoid having to convert it into a string below.

amnn (approved):
LGTM -- one thing I would say is that all your docs so far suggest explicitly naming the schema in cynic's `QueryFragment` macro annotations. This is quite an advanced feature that people only need to worry about if they are using cynic to write custom queries for multiple schemas in the same crate. I think you can avoid mentioning that across all examples, and instead (if you like) adding an addendum somewhere explaining that this is something you could do (or better -- point them to the cynic docs).

bmwill (approved):

PR #58, opened 2024-11-08
[GraphQL Client] Rework errors

stefan-mysten:
Supersedes #45.

### DIFF
--- a/crates/sui-graphql-client/src/error.rs
+++ b/crates/sui-graphql-client/src/error.rs
@@ -12,36 +12,54 @@ use sui_types::types::TypeParseError;
 
 type BoxError = Box<dyn std::error::Error + Send + Sync>;
 
+/// General error type for the client. It is used to wrap all the possible errors that can occur.
 #[derive(Debug)]
 pub struct Error {
     inner: Box<ClientError>,
 }
 
+/// Error type for the client. It is split into multiple fields to allow for more granular error
+/// handling. The `source` field is used to store the original error.
 #[derive(Debug)]
 pub struct ClientError {
+    /// Error when the GraphQL server returns an error.
     pub query_response_error: bool,
+    /// Error when deserialization a value (mostly bcs or base64).
     pub deserialize_error: bool,
+    /// Error when parsing a value.
     pub parse_error: bool,
+    /// Graphql server returned an empty response, though it was expected to return a value.
     pub empty_response_error: bool,
+    /// Error when converting from a type to another.
     pub conversion_error: bool,
+    /// The original error. Use downcasting to get the original error based on the error type from
+    /// above.
     pub source: Option<BoxError>,
 }
 
+/// An empty response with no data from the GraphQL server. This is used to signal that the API
+/// expected some data, but the response was empty.
 #[derive(Debug, Clone)]
 pub struct EmptyResponse;
 
+/// Error when deserialization a value (mostly bcs or base64).
 #[derive(Debug, Clone)]
 pub struct DeserializeError(pub String);
 
+/// Error when parsing a value.
 #[derive(Debug, Clone)]
 pub enum ParseError {
+    /// Error when parsing a URL.
     UrlParseError(String),
+    /// Error when parsing a BigInt into integer.
     BigIntParsingError,
 }
 
+/// Error when converting from a type to another.
 #[derive(Debug, Clone)]
 pub struct ConversionError(pub String);
 
+/// Error when the GraphQL server returns an error.
 #[derive(Debug, Clone)]
 pub struct QueryResponseError(pub Vec<GraphQlError>);
 
@@ -120,12 +138,6 @@ impl From<reqwest::Error> for Error {
     }
 }
 
-impl From<Vec<GraphQlError>> for QueryResponseError {
-    fn from(errors: Vec<GraphQlError>) -> Self {
-        Self(errors)
-    }
-}
-
 impl From<QueryResponseError> for Error {
     fn from(error: QueryResponseError) -> Self {
         Self::empty().with_query_response_error(error)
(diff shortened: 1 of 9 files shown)


crates/sui-graphql-client/src/error.rs
```diff
@@ -0,0 +1,316 @@
...
+    pub deserialize_error: bool,
+    /// Error when parsing a value.
+    pub parse_error: bool,
+    /// Graphql server returned an empty response, though it was expected to return a value.
+    pub empty_response_error: bool,
+    /// Error when converting from a type to another.
+    pub conversion_error: bool,
+    /// The original error. Use downcasting to get the original error based on the error type from
+    /// above.
+    pub source: Option<BoxError>,
```
bmwill, line 37: All these fields need to be private, since the type itself will be private, and accessors should be added to `Error` to get at these inner fields.

crates/sui-graphql-client/src/error.rs
```diff
@@ -0,0 +1,316 @@
...
+/// Error when converting from a type to another.
+#[derive(Debug, Clone)]
+pub struct ConversionError(pub String);
+
+/// Error when the GraphQL server returns an error.
+#[derive(Debug, Clone)]
+pub struct QueryResponseError(pub Vec<GraphQlError>);
+
+#[derive(Debug, Clone)]
+pub struct ConflictingArguments(pub String);
```
bmwill, line 67: I don't think we need each of these error types. Instead we can have an enum `Kind` that can indicate what type of error it is. All source error information should be stored in `source` in its original, un-stringified form. We can either choose to expose the kind and make it public (this would be easiest) so long as its marked as "non-exhaustive" or have "is_XXX" functions for doing those checks ourselves. I think making kind public is probably the easiest thing to do from a maintainability point of view

crates/sui-graphql-client/src/error.rs
```diff
@@ -0,0 +1,316 @@
...
+            result.with_query_response_error(query_response_error.clone())
+        } else if let Some(deserialize_error) = boxed_error.downcast_ref::<DeserializeError>() {
+            result.with_deserialize_error(deserialize_error.clone())
+        } else if let Some(parse_error) = boxed_error.downcast_ref::<ParseError>() {
+            result.with_parse_error(parse_error.clone())
+        } else if let Some(conversion_error) = boxed_error.downcast_ref::<ConversionError>() {
+            result.with_conversion_error(conversion_error.clone())
+        } else {
+            result.with_error(boxed_error)
+        }
```
bmwill, line 249: our library code shouldn't have to do any downcasting, this should be done by users if they want to get the concrete type of the error source

crates/sui-graphql-client/src/lib.rs
```diff
@@ -235,22 +235,22 @@ impl DynamicFieldOutput {
...
+    ) -> Result<T, error::Error> {
         assert_eq!(
             expected_type, &self.name.type_,
             "Expected type {}, but got {}",
             expected_type, &self.name.type_
         );
 
         let bcs = &self.name.bcs;
-        bcs::from_bytes::<T>(bcs).map_err(|_| anyhow!("Cannot decode BCS bytes"))
+        bcs::from_bytes::<T>(bcs).map_err(|e| e.into())
```
bmwill, line 246: For these you should be able to do 
```suggestion
        bcs::from_bytes::<T>(bcs).map_err(Into::into)
```

crates/sui-graphql-client/src/error.rs
bmwill: For the list of GraphQLErrors, since we are tightly coupled to cynic, we can had an Option<Vec<GraphQlError>> field to the Inner Error that can be exposed via a public function `pub fn graphql_errors(&self) -> Option<&[GraphQLError]>`

crates/sui-graphql-client/src/lib.rs
```diff
@@ -376,53 +379,57 @@ impl Client {
...
         if let Some(errors) = response.errors {
-            return Err(Error::msg(format!("{:?}", errors)));
+            return Err(QueryResponseError(errors).into());
         }
 
         response
             .data
             .map(|e| e.chain_identifier)
-            .ok_or_else(|| Error::msg("No data in response"))
+            .ok_or_else(|| EmptyResponse.into())
```
bmwill, line 393: We probably don't need EmptyResponse as a separate error Kind from a Deserialization Error (once we define the error kinds) but may want something like this type or a string that just says the response was empty

crates/sui-graphql-client/src/lib.rs
```diff
@@ -376,53 +379,57 @@ impl Client {
...
+            .map(|x| x.try_into().map_err(|e: ParseError| e.into()))
+            .transpose()
     }
 
     /// Get the protocol configuration.
     pub async fn protocol_config(
         &self,
         version: Option<u64>,
-    ) -> Result<Option<ProtocolConfigs>, Error> {
+    ) -> Result<Option<ProtocolConfigs>, error::Error> {
```
bmwill, line 424: Another option would be to define a Result type and use crate::Result everywhere to avoid needing to specify the Error type in all these functions. You'll still need the error type in scope to do some of these things of course.

crates/sui-graphql-client/src/error.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+impl std::fmt::Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        if let Some(source) = &self.inner.source {
+            writeln!(f, " {}", source)?;
+        }
+        Ok(())
+    }
+}
```
bmwill, line 95: We should probably output the error kind as well?

crates/sui-graphql-client/src/lib.rs
```diff
@@ -548,29 +543,29 @@ impl Client {
...
 
         let total_balance = response
             .data
             .map(|b| b.owner.and_then(|o| o.balance.map(|b| b.total_balance)))
-            .ok_or_else(|| Error::msg("No data in response"))?
+            .ok_or_else(|| {
+                Error::from_error(
+                    Kind::Query,
+                    "Expected response data from query but got None",
+                )
```
bmwill, line 564: Given this does repeat itself frequently it may be worth having a custom constructor for this

stefan-mysten: Yep, better like that!

crates/sui-graphql-client/src/error.rs
stefan-mysten: Fixed

bmwill (approved):

PR #59, opened 2024-11-14
[GraphQL Client] Fix digests

stefan-mysten:
Fix the digests to use the correct types.

### DIFF
--- a/crates/sui-graphql-client/src/lib.rs
+++ b/crates/sui-graphql-client/src/lib.rs
@@ -81,9 +81,9 @@ use streams::stream_paginated_query;
 
 use sui_types::types::framework::Coin;
 use sui_types::types::Address;
+use sui_types::types::CheckpointDigest;
 use sui_types::types::CheckpointSequenceNumber;
 use sui_types::types::CheckpointSummary;
-use sui_types::types::Digest;
 use sui_types::types::Event;
 use sui_types::types::MovePackage;
 use sui_types::types::Object;
@@ -483,7 +483,10 @@ impl Client {
 
     /// The total number of transaction blocks in the network by the end of the provided
     /// checkpoint digest.
-    pub async fn total_transaction_blocks_by_digest(&self, digest: Digest) -> Result<Option<u64>> {
+    pub async fn total_transaction_blocks_by_digest(
+        &self,
+        digest: CheckpointDigest,
+    ) -> Result<Option<u64>> {
         self.internal_total_transaction_blocks(Some(digest.to_string()), None)
             .await
     }
@@ -643,7 +646,7 @@ impl Client {
     /// provided, it will use the last known checkpoint id.
     pub async fn checkpoint(
         &self,
-        digest: Option<Digest>,
+        digest: Option<CheckpointDigest>,
         seq_num: Option<u64>,
     ) -> Result<Option<CheckpointSummary>> {
         if digest.is_some() && seq_num.is_some() {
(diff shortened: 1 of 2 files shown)


crates/sui-graphql-client/src/lib.rs
```diff
@@ -643,7 +646,7 @@ impl Client {
     /// provided, it will use the last known checkpoint id.
     pub async fn checkpoint(
         &self,
-        digest: Option<Digest>,
+        digest: Option<CheckpointContentsDigest>,
```
bmwill, line 649: this probably should be CheckpointDigest and not Contents given the return type here

bmwill: You should take a look at the checkpoint types in the sdk types crate to familiarize yourself for what the differences are. The CheckpointDigest is the digest of a CheckpointSummary while the CheckpointContentsDigest is the digest of the "contents" or the CheckpointContents type which contains the list of txn digests and user signatures

crates/sui-graphql-client/src/lib.rs
```diff
@@ -2126,18 +2133,22 @@ mod tests {
...
         let chckp = client.checkpoint(None, Some(chckp_id)).await;
         assert!(chckp.is_ok());
         let digest = chckp.unwrap().unwrap().content_digest;
+        let chckp_digest = CheckpointDigest::from_bytes(digest.inner()).unwrap();
         let total_transaction_blocks = client
-            .total_transaction_blocks_by_digest(digest.into())
-            .await;
-        assert!(total_transaction_blocks.is_ok());
-        assert!(total_transaction_blocks.unwrap().is_some_and(|tx| tx > 0));
+            .total_transaction_blocks_by_digest(chckp_digest)
```
bmwill, line 2146: This is...odd...and seems inaccurate. total_transaction_blocks_by_digest should take a CheckpointDigest and not use the contents digest.

stefan-mysten: Thanks for catching this. `CheckpointSummary` has a `previous_digest` and `content_digest` field, but no `digest` field which I would expect would return the checkpoint's digest.

stefan-mysten: https://github.com/MystenLabs/sui-rust-sdk/blob/f7921a1354973991a1ef2b3cd9563b8c477c6ef7/crates/sui-sdk-types/src/types/checkpoint.rs?plain=1#L72-L73

crates/sui-graphql-client/src/lib.rs
stefan-mysten: That's what I gathered as well, but I think then we're missing a `digest` field in the `CheckpointSummary` type.

stefan-mysten: Actually, I am confused.

crates/sui-graphql-client/src/lib.rs
stefan-mysten: In GraphQL schema, the `digest` of `Checkpoint` is
```
	"""
	A 32-byte hash that uniquely identifies the checkpoint contents, encoded in Base58. This
	hash can be used to verify checkpoint contents by checking signatures against the committee,
	Hashing contents to match digest, and checking that the previous checkpoint digest matches.
	"""
	digest: String!

```

bmwill: that's because it needs to be calculated, its a hash over the type, its not a field on the actual type. Although graphql likely exposes the digest as a field

stefan-mysten: Got it, thank you! 
Yes, indeed, in GraphQL it's exposed as a field - following where it's being computed, I think this is the code.
https://github.com/stefan-mysten/sui/blob/2cf6f86853604a1ba8c34cc2dc3b54dba47e402f/crates/sui-types/src/messages_checkpoint.rs?plain=1#L194-L201

stefan-mysten:
A proper fix for the `Checkpoint` data is in #62.

PR #62, opened 2024-11-20
[GraphQL Client] Use `bcs` from `Checkpoint` instead of manual conversion.

stefan-mysten:
Get checkpoint summary from bcs instead of manually converting it. This is dependent on first landing
https://github.com/MystenLabs/sui/pull/20340

- [x] Update the schema with the `bcs` comment.
- [x] Update the Sui CLI to enable the new schema support.

Note: this will have to sit for a while because we have no releases in the upcoming weeks so we cannot get an updated version of the CLI that includes the latest schema changes.

### DIFF
--- a/crates/sui-graphql-client/src/lib.rs
+++ b/crates/sui-graphql-client/src/lib.rs
@@ -2133,18 +2133,28 @@ mod tests {
             chckp.unwrap_err()
         );
         let chckp_id = chckp.unwrap().unwrap();
-        let total_transaction_blocks = client.total_transaction_blocks_by_seq_num(chckp_id).await;
-        assert!(total_transaction_blocks.is_ok());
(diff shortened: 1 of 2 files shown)


crates/sui-graphql-client-build/schema.graphql
```diff
@@ -448,6 +448,7 @@ type Checkpoint {
 	By default, the scanning range consists of all transactions in this checkpoint.
 	"""
 	transactionBlocks(first: Int, after: String, last: Int, before: String, filter: TransactionBlockFilter, scanLimit: Int): TransactionBlockConnection!
+	checkpointSummaryBcs: Base64
```
stefan-mysten, line 451: Update the schema when PR in main repo gets merged.

PR #68, opened 2024-12-10
Return a tuple of Event, TxDigest

stefan-mysten:
This PR changes the return type for the `events` API to include the `TransactionDigest` corresponding to the transaction block that issued this event. 

@dario requested this change.

### DIFF
--- a/crates/sui-graphql-client/src/lib.rs
+++ b/crates/sui-graphql-client/src/lib.rs
@@ -940,23 +940,27 @@ impl Client {
         if let Some(events) = response.data {
             let ec = events.events;
             let page_info = ec.page_info;
-            
-            let events_with_digests = ec.nodes.into_iter()
+
+            let events_with_digests = ec
+                .nodes
+                .into_iter()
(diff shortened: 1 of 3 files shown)


crates/sui-graphql-client/src/lib.rs
```diff
@@ -942,15 +942,43 @@ impl Client {
...
+                .iter()
+                .map(|e| e.bcs.0.clone())
                 .map(|b| base64ct::Base64::decode_vec(&b))
                 .collect::<Result<Vec<_>, base64ct::Error>>()?
                 .iter()
                 .map(|b| bcs::from_bytes::<Event>(b))
                 .collect::<Result<Vec<_>, bcs::Error>>()?;
 
-            Ok(Page::new(page_info, nodes))
+            let tx_digests = ec
```
dariorussi, line 953: there is no way to do that in a single loop?
not that it matters much but it feels a bit more clear?

stefan-mysten: Fixed

PR #83, opened 2024-12-19
sui-sdk-types: refactor the types module, folding it into the top-level module

bmwill:


stefan-mysten (approved):
Sweet!

Noting that we'll need to update
- replay
- mvr-cli

PR #87, opened 2025-01-10
feat(sui-sdk-types)!: `Object` content getters

### LINKED ISSUE #72: Contribution: missing `Object` content getters
Hi folks. I've noticed that `sui_sdk_types::types::Object` is missing any way of accessing its inner package/struct contents.

Personally, I'm mostly interested in the BCS bytes of the inner [`MoveStruct`](https://docs.rs/sui-sdk-types/latest/src/sui_sdk_types/types/object.rs.html#190-211). I see that its other attributes are unnecessary from the user's POV, since we already have the `Object::version` and `Object::object_type` methods.

Therefore I'd like to contribute with a new method to `Object`:
```rust
pub fn as_struct(&self) -> Option<(&StructTag, &[u8])>;
```
Naming up for discussion ofc :)

Notice I included the struct tag in the return, otherwise callers run into scenarios like:
```rust
let ObjectType::Struct(tag) = object.object_type() else {
    // ...
};
let bcs = object.as_struct().expect("checked if struct above");
```

unmaykr-aftermath:
- 41a0604 **feat(sui-sdk-types): Object::as_struct getter**

- 9cc7fdd **feat(sui-sdk-types)!: remove `Object::has_public_transfer` getter**
  Probably a good idea to keep `MoveStruct::has_public_transfer` private
  to avoid any misinterpretation there (there's a comment clearly
  stating it's deprecated)

Done as separate commits to make it easy to revert any change.

Closes #72

### DIFF
--- a/crates/sui-sdk-types/src/object.rs
+++ b/crates/sui-sdk-types/src/object.rs
@@ -300,6 +300,13 @@ impl Object {
         }
     }
 
+    pub fn as_struct(&self) -> Option<&MoveStruct> {
+        match &self.data {
+            ObjectData::Struct(struct_) => Some(struct_),
+            _ => None,
+        }
+    }
+
     pub fn owner(&self) -> &Owner {
         &self.owner
     }
--- a/crates/sui-sdk-types/src/object.rs
+++ b/crates/sui-sdk-types/src/object.rs
@@ -223,6 +223,7 @@ impl MoveStruct {
         &self.type_
     }
 
+    #[doc(hidden)]
     pub fn has_public_transfer(&self) -> bool {
         self.has_public_transfer
     }


crates/sui-sdk-types/src/object.rs
```diff
@@ -223,10 +223,6 @@ impl MoveStruct {
         &self.type_
     }
 
-    pub fn has_public_transfer(&self) -> bool {
-        self.has_public_transfer
-    }
-
```
bmwill, line 229: So while the value "has_public_transfer" is deprecated and not used by the system, the value itself is still relevant/important because it impacts the BCS serialization of the object so we still need a way to access that field. We can either add some documentation describing this, or alternatively we could mark it as hidden

bmwill (commented):
Thanks for the quick work here! Aside from the comment on has_public_transfer this looks good!

crates/sui-sdk-types/src/object.rs
unmaykr-aftermath: I would favor `#[doc(hidden)]`, so that we don't risk LSP auto-completion and only people who really looked into the code and are intentional about it will use the method

bmwill: Awesome, would you mind making that change as a part of this PR?

bmwill (approved):

PR #91, opened 2025-01-23
fix: use `owner` instead of `object` in gql `DynamicObjectFieldQuery`

pyncz:
This PR fixes a bug with fetching dynamic _**object**_  fields by changing `object` field in `DynamicObjectFieldQuery` query to `owner`.

I believe that change was already applied for regular DFs, but not object ones, here: #54


stefan-mysten (approved):

stefan-mysten:
Thanks @pyncz!

PR #96, opened 2025-01-30
sui-sdk-types: documentation pass

bmwill:


bmwill:
I think after this lands there is only docs for transaction types left to add

stefan-mysten:
Oh wow, that's a lot of docs. Very cool! I was planning to do some tomorrow, so I will see if I can do some for what's remaining!

PR #101, opened 2025-03-18
[GraphQL] Open up some functions in the graphql

stefan-mysten:
While implementing custom errors, I realized that some functions I'd have to reimplement as they're currently private. This PR opens up most of the required functions (from what I've needed so far).


bmwill:
Looks like there's a clippy lint to fix

PR #102, opened 2025-03-19
Bump cynic to 3.10.0 to match sui repo

stefan-mysten:
Planning to integrate `sui-replay-v2` into main sui repo, so getting some dependencies fixed to be the same.
https://github.com/MystenLabs/sui/pull/21545

### DIFF
--- a/crates/sui-graphql-client-build/Cargo.toml
+++ b/crates/sui-graphql-client-build/Cargo.toml
@@ -4,5 +4,5 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-cynic-codegen = "3.8.0"
+cynic-codegen = "3.10.0"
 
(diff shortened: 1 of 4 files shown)


bmwill:
We can do this but it's not strictly necessary as the Sui repo and select the new version itself

stefan-mysten:
Is that because `sui` monorepo has a newer version? I thought it will fail to build in the `sui` monorepo, because now we have to versions. We used to have it the other way though, here it was 3.8.0 and in `sui` it was 3.7.3

bmwill:
cargo uses semver, 3.10 is a valid version to choose when 3.8 is selected because they are compatible. its only when there are breaking changes (eg 3.10->4.0) that would require that we do a dep update here

PR #103, opened 2025-03-19
Hotfix: add `ConsensusCommitPrologueV4` transaction kind

unmaykr-aftermath:
This `TransactionKind` variant started being produced on testnet epoch 679. However, `sui-sdk-types@0.0.2` doesn't yet have this variant.

Since we were deserializing checkpoints produced by our testnet fullnode into this library's `CheckpointData` type, this led to a `serde` error, which subsequently took down our ingestion pipelines on testnet.

For reference, the error was as follows:
```
Reading checkpoint file /sui/checkpoints/175059402.chk: invalid value: integer '9', expect edvariant index 0 <= i < 9
```

### DIFF
--- a/crates/sui-sdk-types/src/transaction/mod.rs
+++ b/crates/sui-sdk-types/src/transaction/mod.rs
@@ -123,8 +123,47 @@ pub enum TransactionKind {
     ConsensusCommitPrologueV2(ConsensusCommitPrologueV2),
 
     ConsensusCommitPrologueV3(ConsensusCommitPrologueV3),
+    ConsensusCommitPrologueV4(ConsensusCommitPrologueV4),
     // .. more transaction types go here
 }
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde_derive::Serialize, serde_derive::Deserialize)
+)]
+#[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+pub struct ConsensusCommitPrologueV4 {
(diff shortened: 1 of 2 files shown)


unmaykr-aftermath:
@bmwill @stefan-mysten I think it would be more appropriate to merge this into a 'hotfix' branch, from which a patch could be released to [crates.io]. You'll notice this branches off of commit 86a9e06, which is the release of the latest versions on [crates.io]

[crates.io]: https://crates.io/

stefan-mysten:
This already exists on main, we just need to do another release.

unmaykr-aftermath:
> This already exists on main ([0010794](https://github.com/MystenLabs/sui-rust-sdk/commit/00107943f220a6bb9b5f7659344c392a7fea3c45)), we just need to do another release.

Okay. Feel free to close this.

While we're at it, wouldn't it make sense to make `TransactionKind` `#[non_exhaustive]`? So that new variants don't imply breaking changes (well, this would mostly really matter once `sui-sdk-types` is at `0.x.y`)

bmwill:
> While we're at it, wouldn't it make sense to make TransactionKind #[non_exhaustive]? So that new variants don't imply breaking changes (well, this would mostly really matter once sui-sdk-types is at 0.x.y)

Yep it would, there's a number of enums that need #[non_exhaustive] applied to them and we'll do that when we move to 0.x.y versioning (which will happen in a few months once the initial work on stabilizing the new grpc interface is complete).

bmwill:
Pushed out a new release, closing. Thanks @unmaykr-aftermath for flagging this!

PR #104, opened 2025-03-20
Make some mods / types public

0xh3rman:
Hey guys, [we](https://gemwallet.com/) are currently using [sui-types](https://github.com/MystenLabs/sui/tree/main/crates/sui-types) crate from full node repo to build and serialize programmable transactions. 
- Transfer coins
- Split and stake, unstake
- Swap

I believe `sui-rust-sdk` is the best option for this kind of usage, and i tried to replace `sui-types` with `sui-sdk-types` and `sui-transaction-builder` in this [branch](https://github.com/gemwalletcom/core/tree/sui-rust-sdk) to see what's missing.

Here is a short list of things I found along the way
- Some mods and types are better public (this pr shows the change I made)
- Are there any equivalents of these methods in sdk's `TransactionBuilder`? 
- -  `ptb.pay`, `ptb.pay_all_sui`, `ptb.pay_sui`
- - `ptb.pure(amount)`, `ptb.obj()`
- Create `TransactionData` from `Transaction`? there is a `TransactionData::new_programmable` in sui-types
- Equivalent of `IntentMessage::new(Intent::sui_transaction(), tx)` in sdk types?

We're happy to contribute to sdk and accelerate the transition from sui-types to sui-sdk-types

Thanks

### DIFF
--- a/crates/sui-crypto/CHANGELOG.md
+++ b/crates/sui-crypto/CHANGELOG.md
@@ -1,3 +1,7 @@
+# [0.0.3] - 2025-03-20
+
+- Update to use version 0.0.3 of `sui-sdk-types`
+
 # [0.0.2] - 2025-01-06
 
 ## Added
@@ -18,5 +22,6 @@
 
 Initial release
 
+[0.0.3]: https://github.com/mystenlabs/sui-rust-sdk/releases/tag/sui-crypto-0.0.3
 [0.0.2]: https://github.com/mystenlabs/sui-rust-sdk/releases/tag/sui-crypto-0.0.2
 [0.0.1]: https://github.com/mystenlabs/sui-rust-sdk/releases/tag/sui-crypto-0.0.1
--- a/crates/sui-crypto/Cargo.toml
+++ b/crates/sui-crypto/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sui-crypto"
-version = "0.0.2"
+version = "0.0.3"
 authors = ["Brandon Williams <brandon@mystenlabs.com>"]
 repository = "https://github.com/mystenlabs/sui-rust-sdk/"
 license = "Apache-2.0"
@@ -54,7 +54,7 @@ pem = [
 
 [dependencies]
 signature = "2.2"
-sui-sdk-types = { version = "0.0.2", path = "../sui-sdk-types", default-features = false, features = ["hash", "serde"] }
+sui-sdk-types = { version = "0.0.3", path = "../sui-sdk-types", default-features = false, features = ["hash", "serde"] }
 
 # RNG support
 rand_core = { version = "0.6.4", optional = true }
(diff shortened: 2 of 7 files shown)


bmwill:
Happy to learn more about your use cases and learn what's missing so that you can depend on this sdk. Things are still a little wip and things should be coming more together later this year.

As for your pr, non of the things you've marked as public should need to be made public as those are internal details to the library. If you're looking for signing take a look at the sui-crypto crate in this repo.

0xh3rman:
@bmwill thanks for the info, we only need to build and serialize transactions using programmable tx builder, that's exactly we need to know these details (for example `ObjectArg`, `TransactionData` and `IntentMessage`  , we're using another crypto lib to sign txs for all the chains we support.

I agree those constructors are trivial (like `ObjectID::from_single_byte`) and not a blocker, `serialization` is better public IMHO

bmwill:
You're likely looking for the [`Transaction`](https://mystenlabs.github.io/sui-rust-sdk/sui_sdk_types/struct.Transaction.html) type, which is the equivalent of the `TransactionData` type from the sui repo. On the `Transaction` type you'll find the (`Transaction::signing_digest`](https://mystenlabs.github.io/sui-rust-sdk/sui_sdk_types/struct.Transaction.html#method.signing_digest) method which will handle the complexities of calculating the digest needed for signing (no need to worry about an `IntentMessage`). You may also find the [`UserSignature`](https://mystenlabs.github.io/sui-rust-sdk/sui_sdk_types/enum.UserSignature.html) type of interest as this is the signature type (more aptly named than the GenericSignature type in the sui repo).

The internal serialization modules will not be made public as they are internal implementation details of the library, used to implement `serde::Serialize` and `serde::Deserialize` which can be used with the `bcs` library for converting to/from the bcs format. You'll notice that `Transaction` impls [`serde::Serialize`](https://mystenlabs.github.io/sui-rust-sdk/sui_sdk_types/struct.Transaction.html#impl-Serialize-for-Transaction) and [`serde::Deserialize`](https://mystenlabs.github.io/sui-rust-sdk/sui_sdk_types/struct.Transaction.html#impl-Deserialize%3C'de%3E-for-Transaction) so long as the `serde` crate feature is enabled. You can see [here](https://mystenlabs.github.io/sui-rust-sdk/sui_sdk_types/index.html#feature-flags) a list of the feature the crate offers.

We do not yet have a fully fleshed out story for PTB construction in this new sdk but we do have the [`sui-transaction-builder`](https://mystenlabs.github.io/sui-rust-sdk/sui_transaction_builder/index.html) crate which does offer a builder api for constructing user transactions.

0xh3rman:
Thanks for the detail explanation, I did know `serde` feature and `sui-transaction-builder` before asking here, if you don't mind you can take a look at my attempt in this commit: https://github.com/gemwalletcom/core/commit/e1d2567d8640ecc0f0a0fe7edce58eba94f5b6b2

I will try to use `Transaction` type later and see how it goes.

PR #105, opened 2025-03-21
Hotfix: missing `EndOfEpochTransactionKind::StoreExecutionTimeObservations`

unmaykr-aftermath:
This variant is missing in the latest release of `sui-sdk-types`. This time I didn't find it in `master`.

The implementation here is not polished like the rest of the library. See the two TODOs I added. If you want, you can take it over from here, but I'm happy to finish the job if given the right instructions.

IMO, it would be fine to leave the `StoredExecutionTimeObservations` inner type private for now if that means we can get it out faster.

In terms testing, I'm using this branch to patch our `sui-sdk-types` dependency that's used in our testnet ingestion pipelines and everything is working well so far :)

### DIFF
--- a/crates/sui-sdk-types/src/transaction/mod.rs
+++ b/crates/sui-sdk-types/src/transaction/mod.rs
@@ -220,6 +220,7 @@ pub enum TransactionKind {
 ///                                 =/ eoe-deny-list-state-create
 ///                                 =/ eoe-bridge-state-create
 ///                                 =/ eoe-bridge-committee-init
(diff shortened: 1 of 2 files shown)


unmaykr-aftermath:
@bmwill @stefan-mysten any update on the above?

bmwill:
@unmaykr-aftermath Thanks for the ping, Im working on this right now and should be able to get another release out later today

bmwill (approved):

PR #106, opened 2025-03-21
Add single byte constructor to ObjectId

0xh3rman:
Thanks @bmwill for your guidance in draft pr: https://github.com/MystenLabs/sui-rust-sdk/pull/104, I managed to compile our implementation using `sui-sdk-types` and `sui-transaction-builder` 🙏

Will test to see if any functionality breaks later

Can we add 2 simple constructors to build `Address` or `ObjectId` from single byte? this is equivalent to `ObjectID::from_single_byte` in `sui-types`.

### DIFF
--- a/crates/sui-sdk-types/src/address.rs
+++ b/crates/sui-sdk-types/src/address.rs
@@ -70,7 +70,7 @@ impl Address {
         Self(bytes)
     }
 
-    const fn from_u8(byte: u8) -> Self {
+    pub const fn from_u8(byte: u8) -> Self {
         let mut address = Self::ZERO;
         address.0[31] = byte;
         address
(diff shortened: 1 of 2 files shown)


unmaykr-aftermath:
@0xh3rman you may find useful our `const`-ructor for [`ObjectId`](https://docs.rs/af-sui-types/latest/af_sui_types/fn.object_id.html) (there's one for `Address` too). [Here]'s how it's used.

[Here]: https://docs.rs/af-sui-types/latest/src/af_sui_types/sui/transaction/data.rs.html#229-250

0xh3rman:
> @0xh3rman you may find useful our `const`-ructor for [`ObjectId`](https://docs.rs/af-sui-types/latest/af_sui_types/fn.object_id.html) (there's one for `Address` too). [Here]'s how it's used.
> 
> [Here]: https://docs.rs/af-sui-types/latest/src/af_sui_types/sui/transaction/data.rs.html#229-250

But this is another crate?

bmwill:
Just added const functions for constructing Addresses that should obsolesce this pr.

PR #108, opened 2025-05-05
Faucet v2

stefan-mysten:
This PR updates the codebase to use the new v2 faucet APIs.

### DIFF
--- a/crates/sui-transaction-builder/src/lib.rs
+++ b/crates/sui-transaction-builder/src/lib.rs
@@ -482,17 +482,20 @@ mod tests {
     use serde::de;
     use serde::Deserialize;
     use serde::Deserializer;
+
     use sui_crypto::ed25519::Ed25519PrivateKey;
     use sui_crypto::SuiSigner;
-    use sui_graphql_client::faucet::CoinInfo;
     use sui_graphql_client::faucet::FaucetClient;
     use sui_graphql_client::Client;
     use sui_graphql_client::PaginationFilter;
+
+    use sui_types::framework::Coin;
     use sui_types::Address;
     use sui_types::ExecutionStatus;
     use sui_types::IdOperation;
     use sui_types::ObjectId;
     use sui_types::ObjectType;
+    use sui_types::TransactionDigest;
     use sui_types::TransactionEffects;
     use sui_types::TypeTag;
 
@@ -500,7 +503,6 @@ mod tests {
     use crate::Function;
     use crate::Serialized;
     use crate::TransactionBuilder;
-    use sui_types::TransactionDigest;
 
     /// Type corresponding to the output of `sui move build --dump-bytecode-as-base64`
     #[derive(serde::Deserialize, Debug)]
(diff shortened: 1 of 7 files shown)


crates/sui-transaction-builder/src/lib.rs
```diff
@@ -562,35 +564,42 @@ mod tests {
...
-    ///
-    /// NB! This assumes that these tests run on a network whose faucet returns 5 coins per
-    /// each faucet request.
-    async fn helper_setup(
+    async fn helper_setup<'a>(
         tx: &mut TransactionBuilder,
-        client: &Client,
-    ) -> (Address, Ed25519PrivateKey, Vec<CoinInfo>) {
+        client: &'a Client,
+    ) -> (Address, Ed25519PrivateKey, Vec<Coin<'a>>) {
```
bmwill, line 573: I don't think this lifetime is correct as value borrowed wouldn't live long enough. probably more correct to make this 'static

stefan-mysten: Clippy complains when making this static, here (and in all subsequent helper_setup calls)
https://github.com/MystenLabs/sui-rust-sdk/blob/e4d01b452bda913e418165a9490f566c4d3bbdcb/crates/sui-transaction-builder/src/lib.rs?plain=1#L854

bmwill: whats the error you're seeing?

stefan-mysten:
```
838 |         let (address, pk, _) = helper_setup(&mut tx, &client).await;
    |                                ----------------------^^^^^^^-
    |                                |                     |
    |                                |                     borrowed value does not live long enough
    |                                argument requires that `client` is borrowed for `'static`
```

stefan-mysten: If I make `Vec<Coin<'static>>`, then we get this -- note that compiling is fine, but clippy is unhappy
```
error: lifetime may not live long enough
   --> crates/sui-transaction-builder/src/lib.rs:602:9
    |
570 |     async fn helper_setup<'a>(
    |                           -- lifetime `'a` defined here
...
602 |         (address, pk, coins.to_vec())
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'a` must outlive `'static`
 ```

Comment (stefan-mysten) [reply crates/sui-transaction-builder/src/lib.rs:573]: Note, that I am still pretty unfamiliar with lifetime annotations and don't really understand what's going on in this case.

If I remove the `<'a>` lifetime annotation from the function and replace it like this:
```
    async fn helper_setup(
        tx: &mut TransactionBuilder,
        client: &'static Client,
    ) -> (Address, Ed25519PrivateKey, Vec<Coin<'static>>) {
```
 then clippy is complaining.
 ```
 672 |         let client = Client::new_localhost();
    |             ------ binding `client` declared here
673 |         let (_, pk, coins) = helper_setup(&mut tx, &client).await;
    |                              ----------------------^^^^^^^-
    |                              |                     |
    |                              |                     borrowed value does not live long enough
    |                              argument requires that `client` is borrowed for `'static`
    ```

PR #157, opened 2025-09-17
refactor to better capture error info

nickvikeras:
Propagating all of the context necessary to debug transaction execution errors

### DIFF
--- a/crates/sui-rpc/src/client/v2/transaction_execution.rs
+++ b/crates/sui-rpc/src/client/v2/transaction_execution.rs
@@ -1,0 +1,129 @@
+use futures::TryStreamExt;
+use std::time::Duration;
+use tonic::Response;
+
+use crate::client::v2::Client;
+use crate::field::FieldMaskUtil;
+use crate::proto::sui::rpc::v2::ExecuteTransactionRequest;
+use crate::proto::sui::rpc::v2::ExecuteTransactionResponse;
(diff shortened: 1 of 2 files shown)


crates/sui-rpc/src/client/v2/transaction_execution.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// Failed to parse/convert the transaction for digest calculation
+    ProtoConversionError(TryFromProtoError),
+    /// Transaction executed but checkpoint wait timed out
+    CheckpointTimeout(Response<ExecuteTransactionResponse>),
+    /// Transaction executed but checkpoint stream had an error
+    CheckpointStreamError {
+        response: Response<ExecuteTransactionResponse>,
+        error: tonic::Status,
+    },
+}
```
bmwill, line 29: Lets add a `#[non_exhaustive]` to the top of this enum

bmwill: And impl `std::error::Error` for this

nickvikeras: Is the way I implemented it fine? I honestly haven't fully mastered rust error propagation best practices. We don't have anyhow/thiserror/etc in this crate so I just manually implemented the traits.

bmwill: yep lgtm

PR #158, opened 2025-09-18
Adding owned object stream helper

nickvikeras:
I found this to be a bit nicer than dealing with page tokens in a loop. It's similar to the old get_coins_stream function on the old client, but more generic in that it works over objects not coins.

### DIFF
--- a/crates/sui-rpc/src/client/v2/objects.rs
+++ b/crates/sui-rpc/src/client/v2/objects.rs
@@ -1,0 +1,70 @@
+use futures::stream::Stream;
+use futures::stream::{self};
+
+use crate::client::v2::Client;
+use crate::client::v2::Result;
+use crate::proto::sui::rpc::v2::ListOwnedObjectsRequest;
+use crate::proto::sui::rpc::v2::Object;
+
+impl Client {
+    /// Creates a stream of objects based on the provided request.
+    ///
+    /// The stream handles pagination automatically by using the page_token from responses
+    /// to fetch subsequent pages. The original request's page_token is used as the starting point.
+    ///
+    /// # Arguments
+    /// * `request` - The initial `ListOwnedObjectsRequest` with search criteria
+    ///
+    /// # Returns
+    /// A stream that yields `Result<Object>` instances. If any RPC call fails, the
+    /// tonic::Status from that request is returned.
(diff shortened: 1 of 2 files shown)


crates/sui-rpc/src/client/v2/objects.rs
```diff
@@ -0,0 +1,98 @@
...
+        &self,
+        request: impl tonic::IntoRequest<ListOwnedObjectsRequest>,
+    ) -> impl Stream<Item = Result<Object>> + 'static {
+        let client = self.clone();
+        let request = request.into_request();
+
+        // Extract headers and extensions from the original request
+        let headers = request.metadata().clone();
+        let extensions = request.extensions().clone();
+        let request_inner = request.into_inner();
```
bmwill, line 31: We can use [into_parts](https://docs.rs/tonic/latest/tonic/struct.Request.html#method.into_parts) here

crates/sui-rpc/src/client/v2/objects.rs
```diff
@@ -0,0 +1,98 @@
...
+                            request_inner,
+                            client,
+                            headers,
+                            extensions,
+                        ),
+                    ))
+                } else if has_next_page {
+                    let mut new_request = tonic::Request::new(request_inner.clone());
+                    *new_request.metadata_mut() = headers.clone();
+                    *new_request.extensions_mut() = extensions.clone();
```
bmwill, line 58: We can use https://docs.rs/tonic/latest/tonic/struct.Request.html#method.from_parts here

crates/sui-rpc/src/client/v2/objects.rs
```diff
@@ -0,0 +1,98 @@
...
+
+                    match client.state_client().list_owned_objects(new_request).await {
+                        Ok(response) => {
+                            let response = response.into_inner();
+                            let mut data = response.objects;
+                            data.reverse(); // Reverse for efficient pop()
```
bmwill, line 64: Lets change this so that we're currying a https://doc.rust-lang.org/stable/std/vec/struct.IntoIter.html instead of a vec, that will allow us to skip the need to reverse the vector.

bmwill (approved):

PR #234, opened 2026-04-02
sui-rpc: add ByteCount response extension

evan-wall-mysten:
This PR adds a `ByteCount` response extension that counts the number of bytes in the response. It is enabled by calling `with_byte_count()` on the client:
```
let client = Client::new(uri)?.with_byte_count();
let response = client.some_rpc(request).await?;
if let Some(count) = response.extensions().get::<ByteCount>() {
    println!("received {} bytes", count.value());
}
```

It is a replacement for using `prost::Message::encoded_len()` which requires traversing all fields in the message.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 Cargo.lock
+.idea/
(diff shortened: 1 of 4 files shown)


bmwill:
Since we've now added the ability for users to provide arbitrary middleware in https://github.com/MystenLabs/sui-rust-sdk/pull/237, I'm going to close this one as we should be able to leverage that mechanism to get this data.

PR #239, opened 2026-04-10
sui-rpc: shrink execute_transaction_and_wait_for_checkpoint state machine

wbbradley:
Box::pin the inner subscribe_checkpoints, execute_transaction, and get_transaction futures, Box the Streaming handle and the ExecuteTransactionResponse once received, and Box::pin both select! arms. Hoist owned sub-clients so &mut self is not held across awaits, and scope the tonic Request tightly so it drops before the execute await.

The prost-generated ExecutedTransaction inlines Transaction, TransactionEffects, TransactionEvents, ObjectSet and several Vecs of inline submessages, so a single Response<ExecuteTransactionResponse> held across awaits dominates the async state machine. Combined with the subscription Streaming handle and the select! holding both arms inline, the frame blows past 2MB, which is enough to stack-overflow callers whose wrapper future is briefly materialized on a small-stack thread (for example macOS default 2MB) before being moved to the heap via Box::pin. Boxing the large locals and inner futures keeps only single-word pointers in the outer state at each yield point.

Public signature, error variants, and externally observable behavior (subscribe-before-execute ordering, get_transaction fast path, checkpoint/timestamp mutation) are unchanged.

### DIFF
--- a/crates/sui-rpc/src/client/transaction_execution.rs
+++ b/crates/sui-rpc/src/client/transaction_execution.rs
@@ -80,76 +80,86 @@ impl Client {
     /// A `Result` containing the response if the transaction was executed and checkpoint confirmed,
     /// or an error that may still include the response if execution succeeded but checkpoint
     /// confirmation failed.
+    // NOTE: body aggressively Box::pins/Boxes locals and inner futures to shrink the async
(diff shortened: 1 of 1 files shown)


wbbradley:
I'm closing this for now because it might not be needed. I'm getting mixed signals.

PR #241, opened 2026-04-21
[fix] zklogin testnet should use prod verifying key

joyqvq:


crates/sui-crypto/src/zklogin/verify.rs
```diff
@@ -86,7 +86,8 @@ fn bn254_to_fr(f: &Bn254FieldElement) -> Fr {
     Fr::from_be_bytes_mod_order(f.padded())
 }
 
-fn mainnet_verifying_key() -> VerifyingKey {
+/// Used for mainnet only.
```
bmwill: This is inaccurate though right?

crates/sui-crypto/src/zklogin/verify.rs
```diff
@@ -159,7 +160,7 @@ fn mainnet_verifying_key() -> VerifyingKey {
     }
 }
 
-/// Load a fixed verifying key from zkLogin.vkey output. This is based on a local setup and should not use in production.
+/// Load a fixed verifying key from zkLogin.vkey output. This is based on a local setup, used for non-mainnet envs only.
```
bmwill: Same, this is also inaccurate given the fix in the sui repo

bmwill (commented):
So this is a breaking change and so we can't just make this change as-is. Instead you'll want to keep the existing public APIs as-is but mark them with #deprecated and introduce new "prod" ones that the deprecation notice points to, then in the next breaking release we can delete the old apis.

PR #242, opened 2026-04-23
sui-rpc: add from_checkpoint to SubscribeCheckpointsRequest

tpham-mysten:
Adds an optional `from_checkpoint` field to SubscribeCheckpointsRequest so that subscribers can request historical replay starting from a given checkpoint sequence number before continuing with the live stream. When omitted, behavior is unchanged — the stream begins at the latest executed checkpoint.

If the requested checkpoint is no longer available on the server (e.g. outside retention), the server is expected to respond with NOT_FOUND.

This unblocks server-side central backfill in the GraphQL streaming subscription service: on a gRPC disconnect, the server reconnects with `from_checkpoint: last_broadcast + 1` to close the gap without losing events.


bmwill:
This doesn't make sense to add unless we actually support such and interface on the server side. @nickvikeras has talked about possibly doing this but afaik there are no immediate plans for this yet (I could be wrong if you had just recently discussed adding such a feature though)

bmwill:
Ok I got some context from Nick, sounds like we are going to tackle the server side of this now 😀

bmwill:
If possible can you depend on the tip of this pr in a pr in the Sui repo while you're doing the server side implementation? Then when that is ready to land we can land both at the same time. 

I just want to avoid confusion of this not working yet for folks before the server side is done.

tpham-mysten:
I could pin to the tip of this PR while iterating, but I'd prefer to land it now. There'll be a few stacked PRs and it'd be easy to pin one at the wrong commit and ship it.

To avoid the "not working yet" confusion, I just pushed a proto doc note saying from_checkpoint isn't supported yet. I'll also add the guard in `subscription_service.rs`:

```
  if req.from_checkpoint.is_some() {
      return Err(tonic::Status::unimplemented(
          "from_checkpoint is not yet supported",
      ));
  }
```

So clients get a clear error instead of a silent no-op, and the real impl can follow up. WDYT?

tpham-mysten:
As discussed in Slack, server-side resumption in gRPC is probably not the right direction. This is a big architectural decision that's hard to undo, so we want to think about it more carefully before committing.

Quoting @nickvikeras:

> The thing I realized that made me change my mind is that it locks us in to operating a service that can serve both the live stream and historical data. Right now we have that with the full node, but as we scale, if we add a stateless subscription fanout service, we would probably not want it to have to deal with resumption. It won't have a local rocksdb and I don't think we would want to make it proxy requests for missed data.

GraphQL has enough options locally to bridge short gaps without committing to a big architectural change, so closing this for now.

PR #244, opened 2026-05-01
sui-graphql: support custom HTTP headers on Client

### LINKED ISSUE #243: feat(sui-graphql): support custom HTTP headers on Client
`sui-graphql::Client` wraps a default reqwest::Client with no way to attach custom headers to outbound requests currently.

 The sibling `sui-rpc::Client` already supports this via `with_headers(HeadersInterceptor)`. Would be great if `sui-graphql` could expose equivalent functionality, as well.

  I have already implemented the changes in internal fork as this was critical for my usecase. Here is the [commit](https://github.com/phi-labs-ltd/sui-rust-sdk/commit/bee2b2ae5b357cc55368f0129fd2b60076634654).  Happy to open a PR.

abhicris:
Closes #243.

## Summary

`sui-graphql::Client` today wraps a default `reqwest::Client` with no way to attach custom headers, so any caller hitting an authenticated GraphQL gateway (managed providers like Triton/Quicknode, multi-tenant deploys, OTEL-instrumented gateways) has to monkey-patch `reqwest`. The sibling `sui_rpc::Client` already supports this via `HeadersInterceptor`; this PR brings sui-graphql to parity.

API surface (mirrors `sui_rpc::client::HeadersInterceptor` for consistency across the two transports):

```rust
let mut headers = HeaderMap::new();
headers.insert("X-Api-Key", HeaderValue::from_static("..."));
let client = Client::new(Client::MAINNET)?.with_headers(headers);

// or:
let mut client = Client::new(endpoint)?;
client.bearer_auth("eyJ...");
client.basic_auth("alice", Some("hunter2"));
client.extend_headers(extra_headers);
```

`reqwest::header` is re-exported as `sui_graphql::header` so callers don't need a direct `reqwest` dep.

## Tests

Four new wiremock round-trip tests under `client::tests`:

- `with_headers_round_trip` — verifies custom `X-Api-Key` / `X-Tenant` are sent.
- `bearer_auth_round_trip` — verifies `Authorization: Bearer …`.
- `basic_auth_round_trip` — verifies `Authorization: Basic <b64>`.
- `empty_headers_no_op` — proves no auth headers are accidentally injected on a default Client.

```
running 6 tests
test client::tests::test_client_new_invalid_url ... ok
test client::tests::test_client_new ... ok
test client::tests::empty_headers_no_op ... ok
test client::tests::with_headers_round_trip ... ok
test client::tests::basic_auth_round_trip ... ok
test client::tests::bearer_auth_round_trip ... ok
test result: ok. 6 passed; 0 failed
```

## Notes

- Bearer/Basic header values use `set_sensitive(true)` so `reqwest`'s default logger won't print them.
- Basic-auth base64 uses the existing `base64ct` dep — no new deps.
- Pure addition: existing `Client::new` callers compile unchanged.

## Prior art

Used [phi-labs-ltd/sui-rust-sdk@bee2b2a](https://github.com/phi-labs-ltd/sui-rust-sdk/commit/bee2b2ae5b357cc55368f0129fd2b60076634654) as a starting reference. This PR fixes a `HeaderMap` import that pointed at the dev-only `wiremock` crate (would not compile for external consumers), adds `bearer_auth` / `basic_auth` ergonomics matching sui-rpc, and adds the test coverage.

## Test plan

- [x] `cargo build -p sui-graphql`
- [x] `cargo test -p sui-graphql --lib client::tests` (6/6 pass)
- [ ] CI

### DIFF
--- a/crates/sui-graphql/src/client/mod.rs
+++ b/crates/sui-graphql/src/client/mod.rs
@@ -64,7 +64,8 @@ impl Client {
     ///
     /// ```no_run
     /// use sui_graphql::Client;
-    /// use sui_graphql::header::{HeaderMap, HeaderValue};
+    /// use sui_graphql::header::HeaderMap;
+    /// use sui_graphql::header::HeaderValue;
     ///
     /// let mut headers = HeaderMap::new();
     /// headers.insert("X-Api-Key", HeaderValue::from_static("my-key"));
@@ -116,8 +117,7 @@ impl Client {
         let mut header = reqwest::header::HeaderValue::from_str(&value)
             .expect("token is always a valid HeaderValue");
         header.set_sensitive(true);
-        self.headers
-            .insert(reqwest::header::AUTHORIZATION, header);
+        self.headers.insert(reqwest::header::AUTHORIZATION, header);
         self
     }
 
@@ -131,7 +131,8 @@ impl Client {
         U: std::fmt::Display,
         P: std::fmt::Display,
     {
-        use base64ct::{Base64, Encoding};
+        use base64ct::Base64;
+        use base64ct::Encoding;
         let pair = match password {
             Some(p) => format!("{username}:{p}"),
             None => format!("{username}:"),
@@ -140,8 +141,7 @@ impl Client {
         let mut header = reqwest::header::HeaderValue::from_str(&value)
             .expect("base64 is always a valid HeaderValue");
         header.set_sensitive(true);
-        self.headers
-            .insert(reqwest::header::AUTHORIZATION, header);
+        self.headers.insert(reqwest::header::AUTHORIZATION, header);
         self
     }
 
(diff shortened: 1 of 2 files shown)


crates/sui-graphql/src/client/mod.rs
```diff
@@ -48,9 +50,80 @@ impl Client {
...
+        self.headers = headers;
+        self
+    }
+
+    /// Merge the given headers into the client's outgoing-request header set,
+    /// overwriting any existing entries with the same name.
+    pub fn extend_headers(&mut self, headers: HeaderMap) -> &mut Self {
+        self.headers.extend(headers);
+        self
+    }
```
bmwill: The comment isn't exactly correct for what this function does. `extend` will append headers instead of overwriting them. So we either need to update the doc comment or the implementation to match the docs.

abhicris: Good catch — fixed in 1d414f9. I changed the implementation to match the documented intent rather than the doc, since the auth-gateway use case in the PR description (rotating an `X-Api-Key`) actually wants overwrite-on-conflict; `HeaderMap::extend` would have left the stale value alongside the new one.

Behavior now: for each header name in the incoming `HeaderMap`, `insert` on the first value (drops any prior values for that name on the client) and `append` on continuations (preserves multi-value headers from the incoming map). Names not in the incoming map are untouched.

Added `extend_headers_overwrites_existing_keys` round-trip test that asserts `get_all("X-Api-Key").iter().count() == 1` after extension — guards against silent multi-value duplication that wiremock's `header()` matcher wouldn't catch on its own. All 7 `client::tests` pass.

bmwill (approved):

PR #245, opened 2026-05-04
sui-graphql-macros: add graphql_query! for compile-time query validation

tpham-mysten:
## Summary
- Adds `graphql_query!` proc macro in `sui-graphql-macros` that parses and validates GraphQL queries and mutations against the embedded Sui schema using `apollo-compiler`. Unknown fields, undefined variables, and bad arguments fail to compile.
- Migrates all 16 `const QUERY` / `const MUTATION` literals in `crates/sui-graphql/src/client/*.rs` to use the macro.

## Test plan
- `cargo test -p sui-graphql -p sui-graphql-macros`


amnn (approved):
LGTM, thanks @tpham-mysten !

bmwill (approved):

PR #256, opened 2026-05-11
Adding alpha query grpc APIs

nickvikeras:
Adding first iteration of gRPC query APIs.


crates/sui-rpc/proto/sui/rpc/v2alpha/filter.proto
```diff
@@ -0,0 +1,176 @@
...
+    // txn that was mutated/transferred away/deleted/wrapped, or its
+    // address-balance changed via an accumulator event.
+    AffectedAddressFilter affected_address = 2;
+
+    // Match transactions whose effects include a change for the specified
+    // object.
+    AffectedObjectFilter affected_object = 3;
+
+    // Match transactions that made a Move call matching the specified filter.
+    MoveCallFilter move_call = 4;
```
bmwill, line 51: probably worth clarifying that this has to be a top-level entry point, that is nested move calls from inside a move call don't show up here.

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
```diff
@@ -0,0 +1,182 @@
...
+  // Events are returned in ascending or descending packed event sequence order
+  // according to the pagination ordering.
+  rpc ListEvents(ListEventsRequest) returns (stream ListEventsResponse);
+}
+
+// Request message for LedgerService.ListCheckpoints.
+message ListCheckpointsRequest {
+  // Optional. Mask for specifying which parts of the Checkpoint should be
+  // returned (e.g. summary, contents, signatures).
+  optional google.protobuf.FieldMask read_mask = 1;
```
bmwill, line 41: whats the default if not present? Since this is a checkpoint we probably want this to be conservative.

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
```diff
@@ -0,0 +1,182 @@
...
+
+  // Required. Start of the checkpoint range to query (inclusive).
+  optional uint64 start_checkpoint = 2;
+
+  // Required. End of the checkpoint range to query (exclusive). The range
+  // width must not exceed 3,000,000 checkpoints.
+  optional uint64 end_checkpoint = 3;
```
bmwill: do we expect these to be provided? is there a way we can default to some sane values if not provided?

Happy to iterate on this, leaving as-is for now, just want to make sure we're happy with the ergonomics before we stabilize

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// Response message for LedgerService.ListCheckpoints.
+message ListCheckpointsResponse {
+  oneof response {
+    // One matching checkpoint.
+    CheckpointItem item = 1;
+
+    // Final frame indicating the bounded query has no more matching results.
+    EndOfResults end = 2;
```
bmwill: This is where we may want to inject a little bit of information whether there are truly no more items that this list operation can never return, or if the chain progresses you can pass the last cursor and get more entries.

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
```diff
@@ -0,0 +1,182 @@
...
+  // width must not exceed 3,000,000 checkpoints.
+  optional uint64 end_checkpoint = 3;
+
+  // Optional. DNF filter over indexed transaction dimensions. A checkpoint
+  // matches if any transaction it contains satisfies the filter. If absent,
+  // all checkpoints in the range are returned.
+  optional TransactionFilter filter = 4;
+
+  // Optional cursor-bounded pagination. If unspecified, reads in ascending order
+  // with the default page size. The maximum page size is 100; values above
```
bmwill: so if i just provide a default empty value here, what happens at tip of chain?

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
```diff
@@ -0,0 +1,182 @@
...
+    // Final frame indicating the bounded query has no more matching results.
+    EndOfResults end = 2;
+  }
+}
+
+// Request message for LedgerService.ListTransactions.
+message ListTransactionsRequest {
+  // Optional. Mask for specifying which parts of the ExecutedTransaction
+  // should be returned.
+  optional google.protobuf.FieldMask read_mask = 1;
```
bmwill, line 90: same question as above, whats the default?

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
```diff
@@ -0,0 +1,182 @@
...
+
+  // Required. Start of the checkpoint range to query (inclusive).
+  optional uint64 start_checkpoint = 2;
+
+  // Required. End of the checkpoint range to query (exclusive). The range
+  // width must not exceed 3,000,000 checkpoints.
+  optional uint64 end_checkpoint = 3;
```
bmwill: Same question as above

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
```diff
@@ -0,0 +1,182 @@
...
+
+// One event item.
+message EventItem {
+  // Opaque cursor for this item. It may be sent as `pagination.cursor` with
+  // the same request parameters to resume after this item in the current scan
+  // direction.
+  optional bytes cursor = 1;
+
+  // The checkpoint containing the transaction that emitted this event.
+  optional uint64 checkpoint = 2;
```
bmwill, line 167: Do we want this on transactions as well?

bmwill: oh wait. we don't need that since ExecutedTransaction already has this field internally.

crates/sui-rpc/proto/sui/rpc/v2alpha/pagination.proto
```diff
@@ -0,0 +1,36 @@
...
+  // cursor.
+  optional bytes cursor = 2;
+
+  // Ordering for returned results. Defaults to ASCENDING.
+  PaginationOrdering ordering = 4;
+}
+
+// Final response frame indicating the bounded list query has no more matching
+// results. If absent, clients should resume with the last item cursor.
+message EndOfResults {}
```
bmwill: Do we want to include more payload info in this message (similar given my comment above)

bmwill (commented):
Mostly just questions, we don't really need to block landing any of this based on these questions though since we can always iterate while these apis are in alpha

crates/sui-rpc/proto/sui/rpc/v2alpha/filter.proto
```diff
@@ -0,0 +1,176 @@
...
+
+// One event-index predicate.
+message EventPredicate {
+  oneof predicate {
+    // Match events from transactions sent by the specified address.
+    SenderFilter sender = 1;
+
+    // Match events from transactions whose effects include a change for the
+    // specified object.
+    AffectedObjectFilter affected_object = 2;
```
bmwill: This one looks strange, i don't know if we need/want this?

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
nickvikeras: I already did this before I even saw this haha. Yeah it's a good idea.

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
nickvikeras: >oh wait. we don't need that since ExecutedTransaction already has this field internally.

Right

crates/sui-rpc/proto/sui/rpc/v2alpha/pagination.proto
nickvikeras: Yeah definitely, done. I just posted a major update to all of this while you were reviewing it and already did this.

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
nickvikeras: I will document the defaults. It seemed reasonable to just copy them from the get APIs.

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
nickvikeras: The stream ends with an EndOfResults indicating you hit tip of chain. I guess it should be documented.

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
bmwill: Comment from @amnn in graphql cursors are both exclusive worth thinking about if these should be both inclusive or exclusive.

Also what are the semantics of "start" and "end" when going backwards, does start always need to be lower?

nickvikeras: Right now they are required. I suppose you could default to either the starting cursor's checkpoint or 0, and default the END_CHECKPOINT to that plus the max number of checkpoints we use as a db scan bound.

crates/sui-rpc/proto/sui/rpc/v2alpha/query_options.proto
```diff
@@ -0,0 +1,73 @@
...
+
+// Final response frame indicating the effective query interval has no more
+// matching results. The interval is the intersection of checkpoint bounds,
+// cursor bounds, chain/index bounds, and filters. If absent, clients should
+// resume with the last item cursor.
+message EndOfResults {
+  // Boundary that ended the scan. For ascending reads this is the high end of
+  // the effective interval; for descending reads this is the low end.
+  EndOfResultsReason reason = 1;
+}
```
bmwill: So i think many the more important bit of data to return here is a cursor that points to the end of the region that was scanned, that way if we have no hits during an operation, the cursor can still advance.

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
```diff
@@ -40,11 +40,12 @@ message ListCheckpointsRequest {
...
-  // Required. Start of the checkpoint range to query (inclusive).
+  // Optional. Start of the checkpoint range to query (inclusive). Defaults to
+  // genesis.
   optional uint64 start_checkpoint = 2;
 
-  // Required. End of the checkpoint range to query (exclusive). The range
-  // width must not exceed 3,000,000 checkpoints.
+  // Optional. End of the checkpoint range to query (exclusive). Defaults to the
+  // current indexed ledger tip.
   optional uint64 end_checkpoint = 3;
```
bmwill, line 49: given these defaults, what happens if you do a fresh query without any cursor?

crates/sui-rpc/proto/sui/rpc/v2alpha/filter.proto
nickvikeras: Yeah I think that was just another unintended result of me copy-pasting code 3 weeks ago

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
nickvikeras: It emits up to limit_items checkpoints/txns/events starting at genesis, with a bound of 3M checkpoints on the scan. If the scan limit is hit, you get a cursor that lets you resume at checkpoint 3,000,000.

We can change the 3M limit transparently without clients doing anything.

bmwill (approved):

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
amnn: Let me add some more colour on how this is set-up for GraphQL and why:

- We offer three checkpoint bounding filters: `afterCheckpoint`, `atCheckpoint`, `beforeCheckpoint`.
- `after...` and `before...` are **exclusive** bounds to match how cursors work, and to avoid confusion about how the bounds work when working forwards vs backwards.
- For regular pagination APIs, all of these fields are optional.
- For scanning APIs, the plan was that `after...` would be optional, defaulting to starting the range at `0`, but `before...` was going to be required, so that we could always compute the width of the checkpoint range, in a predictable way (i.e. it wouldn't be possible for someone to write a query that worked initially but then eventually failed because the scanned range was too large).
-  To counter this though, we would still have had the regular pagination API which had an "effective" scan range that ended at the latest checkpoint, and extended back 30d or 90d or whatever our configured retention was, because it would be annoying for people to write queries that want to keep up with the tip of the chain otherwise (they would need to initially query the latest checkpoint).

In terms of how this applies to gRPC, it would be good to align gRPC against the query shape from GraphQL for the first two bullets, but the rest is less clear -- we can offer that experience in GraphQL against an API requires explicit bounds always.

Also, sidenote -- ideally the max range would be ~10M or ~30M checkpoints -- matching 30d or 90d retention, roughly.

PR #261, opened 2026-05-17
add support for authenticated events

bmwill:


crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
```diff
@@ -174,6 +174,12 @@ message EventItem {
 
   // The event data.
   optional sui.rpc.v2.Event event = 5;
+
+  // 0-based index of the emitting transaction within its containing
+  // checkpoint. Required for clients verifying authenticated event
+  // streams: this index is part of the BCS-encoded `EventCommitment`
+  // leaf used to construct the per-checkpoint merkle root.
+  optional uint64 transaction_index = 6;
```
bmwill, line 182: cc @nickvikeras adding this here as a placeholder for now. I did validate that its a requirement for the proof machinery so we'll either need to include it here or the client will need to do a checkpoint contents fetch to figure this out themselves.

When we go to stabilize these apis we can reorder the field numbers

PR #264, opened 2026-05-19
Watermarks mid-stream

nickvikeras:
Changing API to allow for watermarks to be returned at arbitrary points in the stream.


crates/sui-rpc/proto/sui/rpc/v2alpha/query_options.proto
```diff
@@ -70,17 +73,45 @@ enum QueryEndReason {
...
+// stream — either standalone or embedded on the last item. Clients have
+// to track the latest watermark themselves anyway (to survive mid-stream
+// disconnects and pre-QueryEnd errors), so QueryEnd carries no cursor of
+// its own: `reason` is the only thing it adds on top of what the stream
+// already delivered.
+message QueryEnd {
   // Reason this response stopped.
-  QueryEndReason reason = 2;
+  QueryEndReason reason = 1;
 }
```
bmwill, line 117: So now that QueryEnd doesn't have a watermark is the assumption that the previous frame will be a watermark frame on query end?

crates/sui-rpc/proto/sui/rpc/v2alpha/query_options.proto
```diff
@@ -70,17 +73,45 @@ enum QueryEndReason {
...
+  // may still hold unscanned matches. Non-decreasing across the stream.
+  //
+  // Unset only on frames whose scan position is still at the genesis
+  // checkpoint (cp 0). Set on every subsequent frame.
+  optional uint64 checkpoint_hi = 2;
+
+  // Descending scans only. Every matching item in checkpoints
+  // `>= checkpoint_lo` has been emitted; checkpoints strictly less may
+  // still hold unscanned matches. Non-increasing across the stream.
+  optional uint64 checkpoint_lo = 3;
```
bmwill, line 102: Fine to keep as-is but i wonder if there is value in having these two fields split out from each other vs having one `checkpoint` field and requiring the client to know what it meant in the context. Obviously this is more explicit.

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
```diff
@@ -148,38 +162,41 @@ message ListEventsRequest {
...
   // streams: this index is part of the BCS-encoded `EventCommitment`
   // leaf used to construct the per-checkpoint merkle root.
-  optional uint64 transaction_index = 6;
+  optional uint64 transaction_index = 5;
+
+  // Progress watermark as of when this item was emitted: its `cursor`
+  // is the resume point past this item, and `checkpoint_hi` /
+  // `checkpoint_lo` (whichever matches the request ordering) is the
+  // inclusive boundary checkpoint that the scan has fully covered.
+  optional Watermark watermark = 6;
```
bmwill: why the change form field 1 to field 6? It may make more sense to have these be anchored in field position 1 but not a big deal.

bmwill (approved):

crates/sui-rpc/proto/sui/rpc/v2alpha/query_options.proto
nickvikeras: No, item frames also carry watermarks, so it could be an item too.

crates/sui-rpc/proto/sui/rpc/v2alpha/ledger_service.proto
nickvikeras: I honestly have no idea. At one point I told claude to stop worrying about the ordering of these because it's all Greenfield and then I guess it decided to move it 🤦‍♂️

PR #282, opened 2026-07-14
[zklogin] support v2 verify vk

joyqvq:


crates/sui-crypto/src/zklogin/verify.rs
```diff
@@ -232,17 +235,104 @@ fn dev_verifying_key() -> VerifyingKey {
     VerifyingKey::new(PreparedVerifyingKey::from(vk))
 }
 
+/// V2 verifying key for production.
+///
+/// TODO: Replace with v2 production verifying key from the ceremony.
+fn mainnet_verifying_key_v2() -> VerifyingKey {
+    mainnet_verifying_key()
+}
```
bmwill: Given we don't have this yet (since we still need to do the ceremony) can we omit this entirely?

crates/sui-crypto/src/zklogin/verify.rs
```diff
@@ -232,17 +235,104 @@ fn dev_verifying_key() -> VerifyingKey {
...
     }
 
-    pub fn new_mainnet() -> Self {
-        mainnet_verifying_key()
+    pub fn new_mainnet_for(version: CircuitVersion) -> Self {
+        match version {
+            CircuitVersion::V1 => mainnet_verifying_key(),
+            CircuitVersion::V2 => mainnet_verifying_key_v2(),
+        }
     }
```
bmwill: Since v2 doesn't exist yet can we skip changing this for now? basically have mainnet construction only support V1 for now instead of "faking" it by exposing an incorrect verifying key for V2.

crates/sui-crypto/src/zklogin/mod.rs
```diff
@@ -7,33 +7,58 @@ use sui_sdk_types::Jwk;
...
+            proof_verifying_keys,
+            circuit_mode: Default::default(),
             jwks: Default::default(),
         }
     }
 
     pub fn new_mainnet() -> Self {
-        Self::new(verify::VerifyingKey::new_mainnet())
+        Self::new(
+            [CircuitVersion::V1, CircuitVersion::V2]
```
bmwill: have this only set V1 for now

bmwill (approved):

PR #293, opened 2026-08-03
sui-sdk-types, sui-rpc: add TransactionExpiration::Validity with allowed proposers

mystenmark:
Adds `TransactionExpiration::Validity`, for transactions that restrict which validators may propose them in consensus, along with its wire representation: a `VALIDITY` kind plus an `AllowedProposers` message carrying committee indices and the epoch those indices refer to.

Depends on MystenLabs/sui-apis#31, which is where the proto change belongs. The vendored copy here is edited directly so the generated code compiles; once the sui-apis PR lands, `make update-protos` should reproduce it exactly.

Note `TransactionExpiration` no longer derives `Copy`, since `AllowedProposers` holds a `Vec`. That is a breaking change for anyone dereferencing one.

Consumed by MystenLabs/sui#27454, which implements the validation and enforcement.

Generated code is from `make proto`. `cargo nextest run --all-features` passes for `sui-sdk-types`, `sui-rpc`, `sui-crypto`, and `sui-transaction-builder` (505 tests), including the BCS roundtrip proptest for the new variant; doctests and `cargo clippy --all-features --all-targets` are clean. The `integration-tests` package fails identically on an unmodified tree — it needs a live network.


crates/sui-sdk-types/src/transaction/mod.rs
```diff
@@ -101,6 +101,73 @@ pub enum TransactionExpiration {
...
+pub struct AllowedProposers {
+    /// The epoch whose committee `proposers` indexes into
+    ///
+    /// Committee indices are only meaningful against one committee, so a set recorded for any
+    /// other epoch is ignored and the transaction is treated as naming no proposers.
+    pub epoch: EpochId,
+    /// Committee indices of the allowed proposers, strictly increasing and non-empty
+    ///
+    /// An empty set is rejected at deserialization, since it names no validator and would be
+    /// rejected on chain.
```
bmwill, line 151: This comment seems to deviate from the documentation listed in the proto definitions. Which is the correct semantics?

crates/sui-sdk-types/src/transaction/mod.rs
```diff
@@ -62,7 +62,7 @@ pub struct SignedTransaction {
 /// transaction-expiration =  %x00      ; none
 ///                        =/ %x01 u64  ; epoch
 /// ```
-#[derive(Clone, Copy, Default, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Default, Debug, PartialEq, Eq, Hash)]
```
bmwill, line 65: This is a breaking change, more a note for myself

mystenmark: oh, hmm. yeah hopefully no one was copying this on its own

crates/sui-sdk-types/src/transaction/mod.rs
mystenmark: The comment is correct - the canonical version uses `NonEmpty` instead of Vec. We could use NonEmpty here if you want to take a new dependency.

claude: update the comments in the proto defs to match this.

bmwill (approved):

bmwill:
@mlogan i think this got stuck due to the github outage last week, can you just rebase it to nudge the ci to run?

PR #298, opened 2026-08-14
feat(graphql): parameterize BigInt on inner type

amnn:
## Description

Previously the SDK assumed all GraphQL `BigInt` values could fit inside a `u64`, but this is not the case. GraphQL uses `BigInt` to represent any integer value that couldn't fit in 53 bits (so doesn't get represented as a JSON number), which means it could be used to represent a `u128` or `i128` (as in the case of a balance change).

This change parameterizes `BigInt` with a sealed inner numeric type that defaults to `u64` for backwards compatibility.

## Test plan

```
$ cargo nextest run -p sui-graphql -- scalars
```


tpham-mysten (approved):
Thanks @amnn!

PR #299, opened 2026-08-14
fix(graphql): prevent value fields from shadowing response input

amnn:
## Description

The `Response` derive generated one local binding per response field. When a field named `value` came before another field, its binding shadowed the `value` parameter containing the original JSON response, so later fields tried to traverse the extracted field instead of the response root, which often led to compilation (type) errors.

Before:

```rust
pub fn from_value(value: serde_json::Value) -> Result<Self, String> {
    let value = {
        let current = &value;
        // extract value
    };
    let digest = {
        let current = &value;
        // incorrectly extracts from the value field
    };
    Ok(Self { value, digest })
}
```

After:

```rust
pub fn from_value(value: serde_json::Value) -> Result<Self, String> {
    Ok(Self {
        value: {
            let current = &value;
            // extract value
        },
        digest: {
            let current = &value;
            // extracts from the original response
        },
    })
}
```

This change initializes fields directly in the struct literal so field names do not introduce local bindings.

## Test Plan

Adds a regression test with a `value` field followed by another extraction.

```
$ cargo nextest run -p sui-graphql-macros --test extraction_test \
  -- test_field_named_value_does_not_shadow_response_value
```

### DIFF
--- a/crates/sui-graphql-macros/src/lib.rs
+++ b/crates/sui-graphql-macros/src/lib.rs
@@ -341,8 +341,7 @@ fn generate_struct_impl(
     let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
 
     // Generate extraction code for each field
-    let mut field_extractions = Vec::new();
-    let mut field_names = Vec::new();
+    let mut field_initializers = vec![];
 
     for field in fields {
         let field_ident = field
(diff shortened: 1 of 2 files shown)


tpham-mysten (approved):
Great catch! Thanks @amnn!

bmwill (approved):

PR #300, opened 2026-08-14
feat(graphql): support flattened response fields

amnn:
## Description

The `Response` derive previously required every struct field to specify a path. This made it awkward to compose reusable response projections that read different fields from the same GraphQL response root.

Add `#[field(flatten)]` for fields that should receive the complete response value instead of a narrowed value:

```rust
#[derive(Response)]
struct ChainInfo {
    #[field(path = "chainIdentifier")]
    chain_id: String,
}

#[derive(Response)]
struct ResponseData {
    #[field(flatten)]
    chain: ChainInfo,
}
```

The derive now generates a borrowed `extract(&serde_json::Value)` method and keeps `from_value(serde_json::Value)` as an owned convenience wrapper. Flattened fields recursively call `extract`, so multiple projections can share the same response without cloning the entire JSON value. GraphQL union responses use the same borrowed extraction path for their inner variants.

`flatten` and `path` are mutually exclusive; combining them produces a compile-time error at the `flatten` attribute.

## Test Plan

Adds extraction coverage for multiple flattened fields receiving the same root value, an `extract`-only field type, and a flattened field named `value`. Adds a compile-fail test for combining `flatten` with `path`.

```
$ cargo nextest run -p sui-graphql-macros
```

### DIFF
--- a/crates/sui-graphql-macros/src/lib.rs
+++ b/crates/sui-graphql-macros/src/lib.rs
@@ -29,7 +29,8 @@
 //!
 //! The macro validates that `object.address` and `object.version` exist in the schema
 //! and that their types match at compile time. It then generates a
-//! `from_value(serde_json::Value) -> Result<Self, String>` method and a `Deserialize`
+//! `from_value(serde_json::Value) -> Result<Self, String>` method, a borrowed
+//! `extract(&serde_json::Value) -> Result<Self, String>` variant, and a `Deserialize`
 //! implementation, so the struct can be used directly with
 //! `serde_json::from_value` or as a response type in GraphQL client calls.
 //!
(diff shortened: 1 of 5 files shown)


tpham-mysten (approved):
Thanks @amnn!

The fragment use case is a good motivation for it. Just a thing that `flatten` is now not compile-time safety. If client defines an incompatible object like `metadata: RandomObject`, it can be failed in runtime. It seems an intended behavior from your test with `ExtractOnly` object.

I'm thinking it would be ideal to enforce a stricter rule on top of this (maybe in a follow-up PR): the type used in `flatten` must declare a `root_type`, and that `root_type` must be compatible with the parent `root_type`.

The idea is that from the parent `root_type` we can compute the list of compatible types from the schema (the type itself, any interface it implements, and any union it belongs to). Then we check which `root_type` the flattened type declares and validate it against that list.

If users want to opt-out to define a custom `ExtractOnly` object, they can use `skip_schema_validation`

I just asked Claude to put up a draft in #303 to show what this would look like. 

WDYT?

bmwill (approved):

PR #302, opened 2026-08-14
refactor(graphql): switch from apollo-compiler to bluejay

amnn:
## Description

Use `bluejay` for GraphQL query parsing and validation in the macro, instead of `apollo-parser`. This cuts down the number of dependencies pulled in by the macro (inclusive) from 74 to 39 (a reduction of almost half -- 35 dependencies removed).

`bluejay` is the GraphQL stack from Shopify. It offers the same functionality modulo error message wording, but with less of a hit on build times.

A reduction of 35 dependencies makes a big difference to a project like jellyfish.

## Test plan

Run existing tests:

```
cargo nextest run -p sui-graphql-macros
```

Check dependency tree:

```
$ cargo tree -p sui-graphql-macros -e normal --prefix none \
     | sed 's/ (\*)$//' \
     | sort -u \
     | wc -l
39
```

### DIFF
--- a/crates/sui-graphql-macros/src/query.rs
+++ b/crates/sui-graphql-macros/src/query.rs
@@ -2,11 +2,10 @@
 //! canonical formatting of GraphQL queries and mutations against the embedded
 //! Sui schema.
 //!
-//! Validation is delegated to `apollo-compiler`, which performs full GraphQL
-//! spec validation (selection set against schema, argument types, fragment
-//! shapes, variable usage, etc.). On success, the macro emits a `&'static str`
-//! literal containing the canonically formatted query. On failure, every
-//! diagnostic becomes a `syn::Error` anchored at the input literal's span.
+//! Parsing, validation, and printing are delegated to Bluejay. On success, the
+//! macro emits a `&'static str` literal containing the canonically formatted
+//! query. On failure, every diagnostic becomes a `syn::Error` anchored at the
+//! input literal's span.
 //!
 //! The macro is intentionally decoupled from any consumer crate: it produces
 //! a plain string literal so callers can wrap it in their own type. The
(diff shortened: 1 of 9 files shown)


crates/sui-graphql-macros/src/query.rs
```diff
@@ -69,35 +93,70 @@ fn expand_impl(input: TokenStream) -> Result<TokenStream2, syn::Error> {
...
+        ExecutableValidator::validate(&document, schema, &cache),
+    ) {
+        return Err(error);
+    }
+
+    let formatted = ExecutableDocumentPrinter::to_string(&document);
+    Ok(quote!(#formatted))
+}
+
+fn query_errors<E>(source: &str, errors: impl IntoIterator<Item = E>) -> syn::Error
```
tpham-mysten: nit: I think we can inline the logic here to avoid having 2 similar function query_errors and combine_query_errors

crates/sui-graphql-macros/tests/compile/graphql_query_invalid_syntax.stderr
```diff
@@ -1,4 +1,4 @@
-error: GraphQL [line 1, col 24]: syntax error: expected R_CURLY, got EOF
```
tpham-mysten, line 1: Did Bluejay hide the error here?

"expected R_CURLY" was a useful message, but I reckon whoever's writing the query should be able to fix it easily enough.

tpham-mysten (approved):
Looking great. Just a few nits. Thanks @amnn

crates/sui-graphql-macros/tests/compile/graphql_query_invalid_syntax.stderr
amnn: Yes, this is variation between Apollo's parser and Bluejay's parser.

crates/sui-graphql-macros/src/query.rs
amnn: Done!

bmwill (approved):

bmwill (approved):

PR #304, opened 2026-08-18
feat(graphql): check flattened projection root types at compile time

tpham-mysten:
Checks at compile time that a flattened field's root type (introduced in #300) is compatible with the root type it is flattened into.

## Problem

Nothing relates a flattened type's `root_type` to the root it is flattened into, so this compiles and fails only at runtime:

```rust
#[derive(Response)]
#[response(root_type = "Epoch")]
struct EpochMetadata { /* reads epochId */ }

#[derive(Response)]
#[response(root_type = "Object")]
struct ObjectResponse {
    #[field(flatten)]
    metadata: EpochMetadata,
}
```

`graphql_query!` rejects the equivalent fragment spread, but `Client::query` takes a plain `&str`, so that only covers queries written through it.

## Rule

A flattened field is extracted unconditionally, so its root must hold for every concrete type the outer position could be: the outer type itself, any interface it implements, or any union it belongs to. For `Object` that is `{Object, Node, IAddressable, IObject}`. The reverse is rejected, since `IObject`'s other implementors do not carry `Object`'s fields.

## How

A proc macro sees the token `EpochMetadata`, not a type, so it cannot look up that type's `root_type`. The check spans two compiler phases: the derive stamps `RESPONSE_ROOT_TYPE` on every response type and emits a `const` block holding the accepted names, which const evaluation compares once rustc has resolved the field's type. Errors are spanned onto the field, nothing reaches the binary, and there are no new dependencies.

A flattened field's type must name the root it reads from. Deriving `Response` supplies it, and a hand-written type can declare `RESPONSE_ROOT_TYPE` alongside its `extract`, which is then accepted at whatever root it names rather than being confined to `Query`. A type supplying neither is told so, rather than being reported as a root type mismatch it has no way to fix.

## Test plan

```
$ cargo test -p sui-graphql-macros
$ RUSTFLAGS="-D warnings" cargo test -p sui-graphql-macros
$ cargo clippy -p sui-graphql-macros --all-targets && cargo fmt --check
```

Unit tests for `find_allowed_flatten_roots`. Runtime tests for interface, union, nested, and mixed flattening. Compile tests for each way a flattened field's type can be wrong: naming no root, providing no `extract`, a defaulted or mismatched `root_type`, a concrete projection flattened into an interface, and a member projection flattened into its union.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017cshe8ffhPkc5cpUS4NLnc


crates/sui-graphql-macros/src/lib.rs
```diff
@@ -412,6 +466,118 @@ fn derive_query_response_impl(input: DeriveInput) -> Result<TokenStream2, syn::E
...
+/// Returns nothing when `field_ty` mentions a generic parameter: the check is a `const`
+/// item, which cannot name generics.
+fn generate_flatten_root_type_check(
+    schema: &schema::Schema,
+    root_type: &str,
+    field_ident: &syn::Ident,
+    field_ty: &syn::Type,
+    generic_params: &[String],
+) -> TokenStream2 {
+    let mentions_generic = field_ty
```
amnn: nit: this seems like quite a low-level way to check for generic types (going down to the token stream and stringifying each token). Is there a more structured way, where you can inspect the various components/variants of a `syn::Type`?

crates/sui-graphql-macros/src/lib.rs
```diff
@@ -412,6 +466,118 @@ fn derive_query_response_impl(input: DeriveInput) -> Result<TokenStream2, syn::E
...
+    // is accepted exactly where `Query` is.
+    let undeclared_ok = roots.contains(&"Query");
+
+    // `str` cannot be compared in a const context, so match the name as bytes.
+    let mut allowed: Vec<_> = roots
+        .iter()
+        .map(|root| syn::LitByteStr::new(root.as_bytes(), field_ty.span()))
+        .collect();
+    if undeclared_ok {
+        allowed.push(syn::LitByteStr::new(UNDECLARED_ROOT, field_ty.span()));
```
amnn: Is this necessary, vs checking whether the root type is `Query` if the field's type has no root type?

crates/sui-graphql-macros/src/lib.rs
```diff
@@ -412,6 +466,118 @@ fn derive_query_response_impl(input: DeriveInput) -> Result<TokenStream2, syn::E
...
+    quote_spanned! { field_ty.span() =>
+        const _: () = {
+            // Marks any field type that did not derive `Response`. An inherent
+            // associated const takes priority over a trait one, so a type that did
+            // resolves to its own declaration, leaving this impl unused.
+            #[allow(dead_code)]
+            trait DefaultRootType {
+                const RESPONSE_ROOT_TYPE: &'static str = #UNDECLARED_ROOT_STR;
+            }
+            impl<T: ?Sized> DefaultRootType for T {}
```
amnn, line 526: This was slightly different from what I had in mind. IIUC this implementation says that if the root type is `Query`, anything can be flattened into it. What I had intended was instead that if you don't specify a root type, `Query` is inferred as the root type, so you would be allowed to do something like this:

```rust
#[derive(Response)]
struct Foo {
  #[field(flatten)]
  chain: Chain
}

#[derive(Response)] 
struct Chain {
  #[field(path = "chainIdentifier")]
  chain_id: Digest,
}
```

Where neither specifies `root_type`, so they both default to `Query`. But not something like this:

```rust
#[derive(Response)]
struct Foo {
  #[field(flatten)]
  chain: Bar
}

struct Bar {
  // Doesn't exist in the schema
  baz: String,
}

impl Bar {
  fn extract(...) { ... }
}
```

crates/sui-graphql-macros/src/schema.rs
```diff
@@ -217,6 +250,84 @@ mod tests {
...
+                "Node"
+            ]
+        );
+    }
+
+    #[test]
+    fn test_find_allowed_flatten_roots_includes_containing_unions() {
+        let schema = Schema::load().unwrap();
+
+        // union DynamicFieldValue = MoveObject | MoveValue
```
amnn, line 281: Is this behaviour for unions safe? It means I can do something like this:

```
#[derive(Response, root_type = "DynamicFieldValue")]
struct DF {
  #[field(flatten)]
  obj: MO,
}

#[derive(Response, root_type = "MoveObject")]
struct MO {
  #[field(path = "content")]
  content: MoveValue,
}
```

Which passes schema validation even though `DF` can fail to parse a `MoveValue`. Is the reason this is okay because we are not strict about marking things as optional?

crates/sui-graphql-macros/src/schema.rs
```diff
@@ -217,6 +250,84 @@ mod tests {
...
+        let roots = schema.find_allowed_flatten_roots("MoveObject");
+        assert!(roots.contains(&"DynamicFieldValue"));
+        assert!(roots.contains(&"MoveObject"));
+        assert!(roots.contains(&"IObject"));
+
+        // A type in no union has no union roots.
```
amnn: This case is kind of strange.

crates/sui-graphql-macros/tests/compile/flatten_root_type_reverse.stderr
```diff
@@ -0,0 +1,5 @@
+error[E0080]: evaluation panicked: `metadata` cannot be flattened into a response rooted at `IObject`: its type must declare `root_type` as one of IObject
```
amnn: This error is a bit confusing. Can we include the actual root_type we detected here? And drop the "one of" if there's only one mentioned type?e


```suggestion
error[E0080]: evaluation panicked: `metadata` cannot be flattened into a response rooted at `IObject`: its type must declare `root_type` as IObject instead of Object.
```

crates/sui-graphql-macros/tests/compile/flatten_root_type_undeclared.stderr
```diff
@@ -0,0 +1,5 @@
+error[E0080]: evaluation panicked: `metadata` cannot be flattened into a response rooted at `Object`: its type must derive `Response`, since a type without the derive can only be flattened into a `Query`-rooted response
```
amnn: Related to my earlier comment, it wasn't my intention to only support hand-written extractors on `Query` root types. Instead of this error message can we say that types must derive `Response` or include a `RESPONSE_ROOT_TYPE` associated constant and `extract` constructor.

amnn: (I also wasn't particularly setting out to support hand-written response types, but it seems like not a bad feature to include at effectively no cost -- it's a good escape hatch).

amnn (commented):
I think I introduced some confusion around my suggestion for supporting that hand-written case for `Query`, I've tried to clarify it in the inline comments.

crates/sui-graphql-macros/src/lib.rs
tpham-mysten: Yeah, I think this wasn't necessary. Claude added it for a generic struct with a flattened field (`inner: T`), where the macro would emit `<T>::RESPONSE_ROOT_TYPE` in a `const` block and hit E0401. But such a struct doesn't compile anyway:

```
    error[E0412]: cannot find type `T` in this scope
    error[E0599]: no function or associated item named `extract` found for type parameter `T`
```

I then dropped it entirely rather than restructuring.

crates/sui-graphql-macros/src/lib.rs
tpham-mysten: Gone now, resolved by your other comment. A type with no declared root isn't accepted anywhere, so there's nothing to add to the allowed list.

crates/sui-graphql-macros/src/lib.rs
tpham-mysten: You're right, I initially assume `Query` was applied to any type without a declared root, hand-written ones included. I have fixed it: a type that declares no root is now rejected outright rather than inheriting `Query`.

crates/sui-graphql-macros/src/schema.rs
tpham-mysten: That direction is already rejected, and now has a test `flatten_member_projection_into_union.rs`:

```
    error[E0080]: `obj` cannot be flattened into a response rooted at `DynamicFieldValue`: its type must declare `root_type` as `DynamicFieldValue`
```

The rule is that the flattened type's root must be the outer root itself, an interface it declares, or a union it belongs to. `DynamicFieldValue` is a union, so only a `DynamicFieldValue`-rooted projection covers it, and `MO` is rooted at one member. That's the case you describe: `DF` could resolve to a `MoveValue` and `MO` reads `contents`, which only `MoveObject` has.

crates/sui-graphql-macros/src/schema.rs
tpham-mysten: Yeah was missing this when reviewing the Claude draft. Removed it

crates/sui-graphql-macros/tests/compile/flatten_root_type_reverse.stderr
tpham-mysten: > ? And drop the "one of" if there's only one mentioned type?

This is done.

> . Can we include the actual root_type we detected here? 

This is unfortunately not possible. The message is a literal baked in when the macro runs, and at that point we only have the token `ObjectMetadata`, not its root. We learn the root at const eval, but there we can't build a string:

```
    error[E0015]: cannot call non-const formatting macro in constants
```

crates/sui-graphql-macros/tests/compile/flatten_root_type_undeclared.stderr
tpham-mysten: Sure. I have updated the logic with new error message

tpham-mysten (commented):
Sorry for the delay @amnn! I have resolved all the comments

PR #311, opened 2026-09-04
sui-rpc: add Client::with_num_connections to spread RPCs over several connections

tpham-mysten:
`sui_rpc::Client` multiplexes every RPC, and every RPC from its clones, over one HTTP/2 connection. That connection's flow-control window and its single driver task cap throughput, so a workload holding many streams in flight can be limited by its own transport before the server is.

```rust
let client = Client::new(uri)?.with_num_connections(4);
```

Opens n connections carrying the same endpoint configuration and assigns each request to the next in rotation. Default stays 1, so existing clients are unchanged; 0 is clamped to 1.

Motivation: the GraphQL server reads kv-rpc through this crate, and subscription backfill issues many concurrent `ListTransactions` calls. Roughly 4 connections per reader lifts backfill toward the per-kv-pod ceiling.

## Notes for review

`Transport` holds the connections, a shared cursor, and the connection bound to the in-flight request, and implements `Service`. `Client::channel()` layers headers, watchdog, request layer and error mapping on top as before. One connection is a one-element rotation, no special case.

- **Assignment is per request, not per service client.** `poll_ready` binds a connection, the paired `call` dispatches to it, which is what lets streams from one held service client occupy different connections. Same shape as `ChannelPool` in `sui-kvstore`, and forced by the tower contract.
- **The bound connection is a `Channel`, not an index.** A ready channel owns a reserved buffer slot, and cloning a channel resets its sender, so dispatching on a different clone panics inside tower.
- The three methods that rebuild the channel now rebuild with the configured count, so reconfiguring no longer collapses the pool.

## Tests

`tests/connection_pool.rs` counts accepted TCP sockets server-side: default opens 1, `with_num_connections(4)` reaches full width then reuses, four concurrently held streams from one service client land on four connections, 0 clamps. Pinning the rotation index to 0 fails the two pool tests.

### DIFF
--- a/crates/sui-rpc/src/client/mod.rs
+++ b/crates/sui-rpc/src/client/mod.rs
@@ -1,4 +1,8 @@
 use std::sync::Arc;
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering;
+use std::task::Context;
+use std::task::Poll;
 use std::time::Duration;
 use tap::Pipe;
 use tonic::body::Body;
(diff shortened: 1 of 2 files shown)


crates/sui-rpc/src/client/mod.rs
```diff
@@ -113,11 +120,100 @@ struct ClientConfig {
...
 
+/// The transport beneath a [`Client`]: a non-empty set of HTTP/2 connections
+/// to one endpoint, with each request assigned to the next connection in
+/// rotation.
+///
+/// A request is bound to a connection by `poll_ready` and dispatched to that
+/// same connection by the `call` that follows, so concurrent RPCs issued
+/// through one service client spread across the set instead of contending for
+/// a single connection's flow-control window and driver task.
+struct Transport {
```
bmwill: I think we should be able to eliminate needing to implement this all ourselves by using the balance channel functionality in `tonic`

nickvikeras: +1 and balance channel implements the p2c load balancing we will want when we turn off client affinity.

tpham-mysten: SG, I have refactored to use balance channel

nickvikeras (approved):

PR #316, opened 2026-09-23
sui-rpc: send x-sui-client-protocol-version request header

mystenmark:
## Description

Send `x-sui-client-protocol-version` on every request: the highest Sui protocol version whose types this SDK can decode (`headers::MAX_PROTOCOL_VERSION`, currently 138). The server can use it to avoid sending types the client can't decode. Headers set with `with_headers` override it.

`MAX_PROTOCOL_VERSION` has to be bumped by hand when the SDK adds support for types from a newer protocol version.

Related PRs:
- Server: https://github.com/MystenLabs/sui/pull/28139
- Rust SDK: https://github.com/MystenLabs/sui-rust-sdk/pull/316
- TypeScript SDK: https://github.com/MystenLabs/ts-sdks/pull/1285

## Test plan

New integration test `tests/client_protocol_version.rs` checks that a mock server receives the header, and that `with_headers` can override it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


crates/sui-rpc/src/headers.rs
```diff
@@ -33,3 +33,14 @@ pub const X_SUI_TIMESTAMP_MS: &str = "x-sui-timestamp-ms";
...
+/// Request header carrying the highest protocol version whose types the client can decode.
+///
+/// The server may use it to avoid sending data the client cannot decode.
+pub const X_SUI_CLIENT_PROTOCOL_VERSION: &str = "x-sui-client-protocol-version";
+
+/// Highest Sui protocol version whose on-chain types this SDK can decode, sent as
+/// [`X_SUI_CLIENT_PROTOCOL_VERSION`].
+///
+/// Bump this when the SDK gains support for types introduced in a newer protocol version.
+pub const MAX_PROTOCOL_VERSION: u64 = 138;
```
mystenmark: double check that this is correct and all messages at this protocol version can be parsed by the sdk

